PR #19, opened 2021-08-16
Coroutines refactoring

mdrlzy:


app/src/main/java/space/taran/arknavigator/mvp/model/repo/PlainTagsStorage.kt
```diff
@@ -32,11 +35,10 @@ class PlainTagsStorage
 
     private var lastModified: FileTime = FileTime.fromMillis(0L)
 
-    private val tagsById: MutableMap<ResourceId, Tags> = {
+    private val tagsById: MutableMap<ResourceId, Tags> = runBlocking(Dispatchers.IO) {
```
kirillt, line 38: Does `runBlocking` block main thread?

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
```diff
@@ -40,7 +40,7 @@ class GalleryPresenter(
         viewState.init(previews)
     }
 
-    fun deleteResource(resource: ResourceId) {
+    fun deleteResource(resource: ResourceId) = presenterScope.launch {
```
kirillt, line 43: Am I right that coroutines launched in `presenterScope` might be cancelled?
For instance, when the presenter is destroyed.

Would this lead to a half-deleted resource in this method?

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
```diff
@@ -56,7 +56,7 @@ class GalleryPresenter(
         return tags
     }
 
-    fun replaceTags(resource: ResourceId, tags: Tags) {
+    fun replaceTags(resource: ResourceId, tags: Tags) = presenterScope.launch {
         Log.d(GALLERY_SCREEN, "tags $tags added to $resource")
         storage.setTags(resource, tags)
```
kirillt, line 61: The same question about `presenterScope`. If this coroutine will be canceled, can it lead to `persist()` call be executed only half-way? If yes, then the storage file might be deleted without new version being written. This might be rare case, but potential outcome is really bad.

app/src/main/java/space/taran/arknavigator/ui/activity/MainActivity.kt
```diff
@@ -32,8 +33,12 @@ class MainActivity : MvpAppCompatActivity(), MainView {
     @Inject
     lateinit var navigatorHolder: NavigatorHolder
 
-    @InjectPresenter
-    lateinit var presenter: MainPresenter
+    private val presenter by moxyPresenter {
```
kirillt, line 36: :+1:

kirillt (commented):
Could you also create some animation (e.g. "endless spinner") when something is being done in io-thread and main threads are idle?

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
mdrlzy: You are right, coroutines launched in `presenterScope` will be canceled if the presenter is destroyed.
I might be wrong, code execution can be stopped when suspend functions are called, because they can throw a `CancellationException` after execution. 

In this case, `storage.remove(resource)` will be fully executed, but this code will be canceled:
```
val path = index.remove(resource)
Log.d(GALLERY_SCREEN, "path $ path removed from index")
Files.delete(path)
```

There are 2 possible solutions:
- Use a different scope for important work. For example `appScope`, which will be tied to the app lifecycle. 
- Mark important work as `presenterScope.launch(NonCancellable)`

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
mdrlzy: Yes, `persist()` can be half done if it contains multiple suspend functions inside. Similar to `replaceTags(resource: ResourceId, tags: Tags)` and `storage.setTags(resource, tags)`

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PlainTagsStorage.kt
mdrlzy: Yes, it does. Dispatchers.IO is useless)

It must be replaced with suspend constructor.

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
kirillt: `NonCancellable` sounds as easier way, what are pros/cons of these 2 options?

Why `storage.remove(resource)` will be fully executed? Is it because it's coroutine is in another scope?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PlainResourcesIndex.kt
```diff
@@ -138,21 +136,21 @@ class PlainResourcesIndex internal constructor (
...
         //todo: parallel and asynchronous
-        internal fun scanResources(files: List<Path>): Map<Path, ResourceMeta> =
+        internal suspend fun scanResources(files: List<Path>): Map<Path, ResourceMeta> = withContext(Dispatchers.IO) {
             files.map {
                 it to ResourceMeta(
                     id = computeId(it),
-                    modified = Files.getLastModifiedTime(it))
-            }
-            .toMap()
+                    modified = Files.getLastModifiedTime(it)
```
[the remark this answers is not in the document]
mdrlzy: I think this is an ide inaccuracy. The inspection cannot determine which thread will be blocked.
Coroutine runs on IO thread, where blocking is ok.
https://youtrack.jetbrains.com/issue/KTIJ-838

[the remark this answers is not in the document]
kirillt: Got it, thanks for the link

app/src/main/java/space/taran/arknavigator/ui/fragments/FoldersFragment.kt
```diff
@@ -11,62 +11,37 @@ import androidx.recyclerview.widget.GridLayoutManager
...
-    private lateinit var roots: Set<Path>
-    private lateinit var favorites: Set<Path>
-
     @Inject
     lateinit var router: Router
 
-    @InjectPresenter
-    lateinit var presenter: FoldersPresenter
+    private lateinit var foldersTree: FoldersTree
+    private var folderPicker: FolderPicker? = null
```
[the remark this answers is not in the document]
mdrlzy: Because after screen rotation, `updateRootPickerDialogPath(path: Path)` can be called via moxy before `FolderPicker` initialization.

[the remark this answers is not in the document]
mdrlzy: Use case:
open the dialog -> click on the folder -> close the dialog -> open the dialog again -> rotate the screen

`setRootPickerDialogVisibility(devices)` -> `updateRootPickerDialogPath(path)` -> `setRootPickerDialogVisibility(null)` -> `setRootPickerDialogVisibility(devices)`

After screen rotation the moxy will restore the state using AddToEndSingleStrategy:
`updateRootPickerDialogPath(path)` -> `setRootPickerDialogVisibility(devices)`

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
mdrlzy: I have too little experience with coroutines, I can't say the pros and cons. Google recommends using a different scope.
https://developer.android.com/kotlin/coroutines/coroutines-best-practices#create-coroutines-data-layer

PR #28, opened 2021-08-19
Draggable tags selector

mdrlzy:


app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -53,6 +57,13 @@ class ResourcesFragment(val root: Path?, val path: Path?): MvpAppCompatFragment(
     private var sorting: Sorting = Sorting.DEFAULT
     private var ascending: Boolean = true
 
+    private val rootLayoutY by lazy {
```
kirillt, line 60: Maybe `layoutBottom`?

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -53,6 +57,13 @@ class ResourcesFragment(val root: Path?, val path: Path?): MvpAppCompatFragment(
...
+        val loc = IntArray(2)
+        layout_root.getLocationOnScreen(loc)
+        loc[1]
+    }
+    private val rootLayoutHeight by lazy { layout_root.height }
```
kirillt, line 65: Maybe `layoutHeight`?

kirillt (commented):
Looks good. What would be also nice to add:
1) When we go into gallery mode and back — we should restore previous height of the selector.
2) Acceleration near edges would be nice (so a user could swipe quickly for 100% height or, say, 20%).

Issue #3, opened 2021-08-11
Make all interactions with Room non-blocking


kirillt:

kirillt:
Implemented in #30

PR #30, opened 2021-08-19
Coroutines refactoring without Folders refactoring

mdrlzy:


app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
```diff
@@ -40,7 +40,7 @@ class GalleryPresenter(
         viewState.init(previews)
     }
 
-    fun deleteResource(resource: ResourceId) {
+    fun deleteResource(resource: ResourceId) = presenterScope.launch {
```
kirillt, line 43: Regarding potential cancellation of coroutines and ways to solve it:
I think `NonCancellable` is essential here, since it guarantees that the coroutine will not be cancelled at all.
All operations writing to tags storage (filesystem) must be not possible to cancel.

Operations writing to index (Room) are less important, because right now it is just cache.
But when we implement more advanced features like version tracking, the index will become more important.
We can make these operations `NonCancellable` as well.

kirillt (commented):
Could you also create some animation (e.g. "endless spinner") when something is being done in io-thread and main threads are idle?

kirillt:
Closes #3

Issue #20, opened 2021-08-16
ResourcesFragment: "tags off" gets stucked


kirillt:
How to reproduce:
1. open a folder
2. click "tags off" button (picture of crossed-out cloud)
3. undo by clicking "tags on"
4. open any resource
5. go back (using system navigation button)

Expected:
* all resources in the folder displayed

Actual:
* only resources without tags displayed (the same as after step 2)

kirillt:
Fixed in #37

Issue #12, opened 2021-08-11
Resources grid forgets sorting after tags changed in the selector


kirillt:
1. Open a folder.
2. [optionally] select some tags
3. Choose sorting type. Now resources are sorted correctly.
4. Select different tags so another resources are presented.

Expected: resources are sorted with ordering type specified before.
Actual: resources are not sorted. Sorting menu displays ordering type chosen before.

kirillt:
Fixed

PR #37, opened 2021-08-29
Resources refactoring

mdrlzy:
See #12 #20


app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/TagsSelectorPresenter.kt
```diff
@@ -0,0 +1,145 @@
...
+    fun onClearClick() {
+        (included + excluded).forEach { uncheckTag(it, needToCalculate = false) }
+        calculateTagsAndSelection()
+    }
+
+    fun calculateTagsAndSelection() {
+        if (storage == null || resources == null)
+            return
+
+        val allTags = storage!!.listAllTags()
```
kirillt, line 61: This is not equivalent to `resources.flatMap { storage!!.getTags(it) }` (see other comment about `listAllTags`). Maybe we could design `TagsStorage` as more granular, so we would create one tags storage per every favorite, but I am not sure that we would benefit from it.

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/TagsSelectorPresenter.kt
```diff
@@ -0,0 +1,145 @@
...
+
+        excluded = excluded.intersect(allTags).toMutableSet()
+        included = included.intersect(allTags).toMutableSet()
+
+        val selectionAndComplementWithTags = resources!!
```
kirillt, line 66: We iterate `resources` anyway. Why not just take values of this intermediate map, concatenate them and use instead of `allTags`?

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/TagsSelectorPresenter.kt
```diff
@@ -0,0 +1,145 @@
...
+        calculateTagsAndSelection()
+    }
+
+    fun calculateTagsAndSelection() {
+        if (storage == null || resources == null)
+            return
+
+        val allTags = storage!!.listAllTags()
+
+        excluded = excluded.intersect(allTags).toMutableSet()
```
kirillt, line 63: How can `excluded` or `included` be bigger than `allTags`?

app/src/main/java/space/taran/arknavigator/ui/fragments/GalleryFragment.kt
```diff
@@ -37,7 +36,7 @@ import space.taran.arknavigator.utils.*
 class GalleryFragment(
     private val index: ResourcesIndex,
     private val storage: TagsStorage,
-    private val resources: ResourcesList,
+    private val resources: MutableList<ResourceId>,
```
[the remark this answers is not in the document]
mdrlzy: To keep the data up to date in case of screen rotation, you just need to store it in the presenter.

`ResourceGripPresenter` does not know where to get resources, they are provided by the `TagSelectorPresenter` or `ResourcesPresenter`. Now both of them are using the `ResourceIndex`. Since both resource and gallery screens refer to the same `index` (object) and it is stored in the presenter, there is no need to worry about transferring changes between screens and data persistence in case of rotating screens.

But if the app is killed by the Android and the user returns to the app, then the data will be restored from the bundle (if we switch to them). The data from the bundle will most likely be no longer up-to-date and it will be necessary to check the list of transferred resources for changes.

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
```diff
@@ -0,0 +1,98 @@
...
+        this.sorting = sorting
+        pushResources()
+    }
+
+    fun updateAscending(ascending: Boolean) {
+        this.ascending = ascending
+        pushResources()
+    }
+
+    private fun pushResources() {
```
kirillt, line 86: This name is still confusing :) Maybe just `sortResources`?
"Push" sounds like there is a collection which grows in size with time.

kirillt (approved):
Good job.

Issue #35, opened 2021-08-27
Persisted sorting preferences


kirillt:
Sorting preferences must be saved between runs. E.g. if a user chooses "sort by size, descending", this choice must be still enabled even after the user left the app, or the app crashed and restarted.

CloudLevi:
Sorting preferences are now persistent.
Also, fixed an a bug where the app would crash when "Extension" sorting is selected (The app wasn't checking whether the current file name has an extension or not).

kirillt:
Implemented in #62

PR #62, opened 2021-09-06
#35 Persisted sorting preferences

CloudLevi:
- Sorting preferences are now persistent.
- Fixed bug when clicking extension sort mode


app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -76,6 +78,11 @@ class ResourcesFragment(val root: Path?, val path: Path?): MvpAppCompatFragment(
         savedInstanceState: Bundle?): View? {
 
         Log.d(RESOURCES_SCREEN, "inflating layout for ResourcesFragment")
+        sharedPrefs = requireContext().getSharedPreferences("user_preferences", Context.MODE_PRIVATE)
+
+        sorting = Sorting.values()[sharedPrefs.getInt("sorting_preference", 0)]
```
kirillt, line 83: Separate variable for preference would make code more readable.

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -261,6 +266,20 @@ class ResourcesFragment(val root: Path?, val path: Path?): MvpAppCompatFragment(
         dialog = alertBuilder.show()
     }
 
+    private fun sortingCategorySelected(itemID: Int): Sorting{
+        return when(itemID) {
+            R.id.rb_default -> {
+                Sorting.DEFAULT
+                throw AssertionError("As-is sorting is initial, unsorted order")
```
kirillt, line 273: This is interesting case. Sometimes, a user might want "as-is" sorting, because it is slightly faster and in some cases order doesn't matter. E.g. later we will have "shuffle" button, so we don't need to sort elements if the user wants to use it. But if we change sorting and remember it, then we can't switch to "as-is". So, we should make it selectable, but since it doesn't make sense to "unsort" resources it should be active only starting next time.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -94,8 +94,13 @@ fun findLongestCommonPrefix(paths: List<Path>): Path {
...
-        .split('.')
-        .drop(1)
-        .last()
\ No newline at end of file
+fun extension(path: Path): String {
+    val pathString = path.fileName.toString()
+    return if (pathString.contains(".") &&
+        !pathString.split(".").drop(1).isNullOrEmpty())
+        pathString
+            .split('.')
```
kirillt, line 102: I think we don't need several calls to searching functions.
Maybe something like this? https://tinyurl.com/we3wrzm2

kirillt (commented):
Good work, but fix couple of small things please.

mdrlzy:
Work with `SharedPreferences` needs to be wrapped in `UserPreferences(val context: Context)`, and it should be delivered to a fragment via an inject. Can you also add `ascending` to the `UserPreferences`?

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
mdrlzy: Yes, it would be better to create constants for preference names.

mdrlzy (commented):
Work with `SharedPreferences` needs to be wrapped in `UserPreferences(val context: Context)`, and it needs to be delivered to a fragment via an inject.
Can you also add `asceding` to the `UserPreferences`?

app/src/main/java/space/taran/arknavigator/utils/DataStoreManager.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.datastore.preferences.core.booleanPreferencesKey
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.intPreferencesKey
+import androidx.datastore.preferences.preferencesDataStore
+import kotlinx.coroutines.flow.first
+import javax.inject.Inject
+
+const val SHARED_PREFERENCES_KEY = "user_preferences"
+
+private val Context.preferencesDatastore by preferencesDataStore((SHARED_PREFERENCES_KEY))
```
mdrlzy, line 13: `SHARED_PREFERENCES_KEY` and `by preferencesDataStore((SHARED_PREFERENCES_KEY))` must be inside the `DataStoreManager` because `DateStore` will not be used without a `DataStoreManager`

app/src/main/java/space/taran/arknavigator/utils/DataStoreManager.kt
```diff
@@ -0,0 +1,36 @@
...
+
+class DataStoreManager @Inject constructor(val context: Context){
```
mdrlzy, line 15: Can you rename to `UserPreferences` for clarity and move to mvp/model?

app/src/main/java/space/taran/arknavigator/utils/DataStoreManager.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    suspend fun getUserSorting(): Int =
+        dataStore.data.first()[PreferencesKeys.SORTING_PREFERENCE] ?: 0
+
+    suspend fun setUserSortOrder(isAscending: Boolean){
+        dataStore.edit { preferences -> preferences[PreferencesKeys.SORTING_ORDER] = isAscending}
+    }
+
+    suspend fun getUserSortOrder(): Boolean =
+        dataStore.data.first()[PreferencesKeys.SORTING_ORDER] ?: false
```
mdrlzy, line 30: Maybe `isUserSortAscending` is better?
Should the default value be `false`?

app/src/main/java/space/taran/arknavigator/extensions/ViewExt.kt
```diff
@@ -0,0 +1,9 @@
+package space.taran.arknavigator.extensions
+
+import android.view.View
+
+fun View.changeEnabledStatus(isEnabledStatus: Boolean){
+    isEnabled = isEnabledStatus
+    isClickable = isEnabledStatus
+    isFocusable = isEnabledStatus
+}
```
mdrlzy, line 9: I think the `extensions` package should be inside `utils`

app/src/main/java/space/taran/arknavigator/di/modules/AppModule.kt
```diff
@@ -1,15 +1,30 @@
...
 import dagger.Module
 import dagger.Provides
+import space.taran.arknavigator.ui.App
+import space.taran.arknavigator.utils.DataStoreManager
+import javax.inject.Singleton
+
 
 @Module
-class AppModule(val app: App) {
+class AppModule(val app: App, val context: Context) {
```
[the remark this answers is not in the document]
CloudLevi: Sure, I haven't thought about it in the moment for some reason.

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
```diff
@@ -30,9 +32,24 @@ class ResourcesPresenter(
...
+    @Inject
+    lateinit var dataStoreManager: DataStoreManager
+
     private lateinit var index: ResourcesIndex
     private lateinit var storage: TagsStorage
 
+    var sorting: Sorting = Sorting.DEFAULT
+        set(value) {
+            field = value
+            presenterScope.launch { dataStoreManager.setUserSorting(value.ordinal) }
```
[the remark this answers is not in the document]
CloudLevi: I've experimented with different ways of setting these values, and this one seemed the best, since it takes up less space, and moving away from overpopulating the class with getters/setters like it was in Java, but no prob, I'll replace these with dedicated methods.

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
```diff
@@ -48,6 +65,12 @@ class ResourcesPresenter(
         return TagsSelector(tags, resources().toSet(), storage)
     }
 
+    private fun getSorting() = presenterScope.launch {
+        sorting = Sorting.values()[dataStoreManager.getUserSorting()]
```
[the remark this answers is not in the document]
CloudLevi: The whole point of `getSorting()` method is to update presenter values with the values from Datastore and notigy UI about the updated values. It's an asynchronous procedure, so returning the values it to the UI isn't really possible. Access to the values is direct. Perhaps, renaming the method to `fetchSorting()` would make it better to clear up any confusions?

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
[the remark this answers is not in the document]
kirillt: If @mdrlzy means separate dedicated setter, then yeah, maybe explicit function would be clearer that something complex is going on inside.

app/src/main/java/space/taran/arknavigator/utils/DataStoreManager.kt
CloudLevi: The reason behind naming it DatastoreManager was to have a distinct understanding, that it's indeed Datastore, and not a SharedPreferences helper class etc.  I'll rename it to UserPreferences and move it.

app/src/main/java/space/taran/arknavigator/utils/DataStoreManager.kt
kirillt: Right, boolean things should be explicit about what is `true` or `false`.

app/build.gradle
```diff
@@ -71,6 +71,9 @@ dependencies {
 
     implementation "ru.terrakok.cicerone:cicerone:5.0.0"
 
+    // DataStore
```
kirillt, line 74: No need for this comment :)

app/src/main/java/space/taran/arknavigator/utils/DataStoreManager.kt
CloudLevi: Renamed. Also, of course, default should be true. Changed.

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -261,6 +252,33 @@ class ResourcesFragment(val root: Path?, val path: Path?): MvpAppCompatFragment(
...
+            Sorting.NAME -> gridAdapter.sortBy { it.fileName }
+            Sorting.SIZE -> gridAdapter.sortBy { Files.size(it) }
+            Sorting.TYPE -> gridAdapter.sortBy { extension(it) }
+            Sorting.LAST_MODIFIED -> gridAdapter.sortBy { Files.getLastModifiedTime(it) }
+            Sorting.DEFAULT -> {
+                sendShortToastMessage(requireActivity().getString(R.string.as_is_sorting_selected))
+            }
+        }
+
+        Log.d("TAG", "sortAccordingToChoice: sorting: ${presenter.sorting}, reverse = ${presenter.sortOrderAscending}")
```
kirillt, line 266: I guess you forgot to replace `"TAG"` with `RESOURCES_SCREEN`

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -312,6 +330,17 @@ class ResourcesFragment(val root: Path?, val path: Path?): MvpAppCompatFragment(
         updateVerticalBias(iv_drag_handler)
     }
 
+    private fun sendShortToastMessage(message: String){
```
kirillt, line 333: Actually, there is `notifyUser` method in `Notifications` file.

app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,4 @@
 <resources>
     <string name="app_name">ARK Navigator</string>
+    <string name="as_is_sorting_selected">This option will be applied when the screen is re-opened.</string>
```
kirillt, line 3: Don't want to be picky, but I think dots at the end are not necessary in one-line messages.
Also, maybe "This option will be applied during next resources selection" would be more semantically correct.

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -261,6 +252,33 @@ class ResourcesFragment(val root: Path?, val path: Path?): MvpAppCompatFragment(
...
     }
 
+    private fun sortAccordingToChoice(){
+        when(presenter.sorting) {
+            Sorting.NAME -> gridAdapter.sortBy { it.fileName }
+            Sorting.SIZE -> gridAdapter.sortBy { Files.size(it) }
+            Sorting.TYPE -> gridAdapter.sortBy { extension(it) }
+            Sorting.LAST_MODIFIED -> gridAdapter.sortBy { Files.getLastModifiedTime(it) }
+            Sorting.DEFAULT -> {
+                sendShortToastMessage(requireActivity().getString(R.string.as_is_sorting_selected))
```
kirillt, line 262: This message about `as-is` pops up every time a user opens the resources screen, not only after making the choice itself.

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -312,6 +330,17 @@ class ResourcesFragment(val root: Path?, val path: Path?): MvpAppCompatFragment(
         updateVerticalBias(iv_drag_handler)
     }
 
+    private fun sendShortToastMessage(message: String){
+        Toast.makeText(requireContext(), message, Toast.LENGTH_SHORT).show()
+    }
+
+    private fun changeSortOrderLockStatus(dialogView: View, isEnabledStatus: Boolean){
```
kirillt, line 337: Just a question: what is "lock" here?

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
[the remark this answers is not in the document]
mdrlzy: I meant `dataStoreManager.getUserSorting()` should return `Sorting` not `Int`.
`Sorting.values()..` must be inside.

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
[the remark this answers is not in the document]
mdrlzy: I meant `dataStoreManager.setUserSorting` must accept `Sorting` not `Int`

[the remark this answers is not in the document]
CloudLevi: Oh. No problem. Makes total sense.

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
CloudLevi: I guess I wasn't able to come up with an elegant explanation of the method in its name :)
`changeSortOrderLockStatus` enables or disables "Ascending" and "Descending" radio buttons according to the passed value. I just wrapped it in one method for convenience.

app/src/main/java/space/taran/arknavigator/ui/fragments/GalleryFragment.kt
```diff
@@ -196,22 +202,24 @@ class GalleryFragment(
 
         val tags = presenter.listTags(resource)
 
-        dialogView = LayoutInflater.from(requireContext()).inflate(R.layout.dialog_tags, null)
-        val alertDialogBuilder = AlertDialog.Builder(requireContext()).setView(dialogView)
+        //dialogView = LayoutInflater.from(requireContext()).inflate(R.layout.dialog_tags, null)
```
kirillt, line 205: Commented-out code

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
```diff
@@ -0,0 +1,42 @@
...
+import javax.inject.Inject
+
+class UserPreferences @Inject constructor(val context: Context) {
+    private val SHARED_PREFERENCES_KEY = "user_preferences"
+
+    private val Context.preferencesDatastore by preferencesDataStore((SHARED_PREFERENCES_KEY))
+
+    private val dataStore = context.preferencesDatastore
+
+    suspend fun setUserSorting(sorting: Sorting) {
```
kirillt, line 20: Just `setSorting` and `setSortingAscending` should be enough. There is already `User` word in the name of the class, so no confusion possible.

app/src/main/java/space/taran/arknavigator/ui/fragments/FoldersFragment.kt
```diff
@@ -191,34 +197,36 @@ class FoldersFragment: MvpAppCompatFragment(), FoldersView, BackButtonListener {
...
-                    dialogView.btn_roots_dialog_pick.text = PICK_ROOT
-                    rootNotFavorite = true
+        dialogBinding.apply {
+            if (Files.isDirectory(path)) {
+                folderPicker.updatePath(path)
+
+                val rootPrefix = roots.find { path.startsWith(it) }
+                if (rootPrefix != null) {
+                    if (rootPrefix == path) {
+                        //todo fake disabling (still show messages when pressing on disabled button)
```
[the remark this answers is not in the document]
kirillt: I'm wrong, they are still there, I'll remove them myself

kirillt (approved):

kirillt:
Closes #35

Issue #45, opened 2021-09-01
"Edit resource" button


kirillt:
In gallery screen, a resource must be capable of edition. The "edit tags" button should be moved into tags list so its place could be used by new "edit resource" button.

PR #68, opened 2021-09-10
#6 "open with" button in gallery mode

CloudLevi:
Issues covered:
#6 "open with" button in gallery mode
#45 Edit resource button


app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -48,6 +63,9 @@ fun isHidden(path: Path): Boolean =
 fun folderExists(path: Path): Boolean =
     Files.exists(path) || Files.isDirectory(path)
 
+fun isFileEmpty(filePath: Path?): Boolean =
```
kirillt, line 66: I think this function is little bit harmful. We should not tolerate null paths, if we encounter them in `GalleryFragment` previews, then it is a mistake and we have to throw `AssertionError()`.

kirillt (commented):
Could you also put "pencil" icon on edit button for images?

Issue #95, opened 2021-10-08
"Link to Web" resource kind


kirillt:
It should be possible to work with URLs using the app.
This allows a user to use our app as bookmarks keeper.

1. Sharing a link from external apps must result in saving the link into one of user's folders.
2. Links should be represented as separate entities in resources grid and gallery view.
3. Opening any link must be possible with Firefox or Chrome.
4. Opening specific links should be possible with specialized apps (e.g. YouTube, MeetUp).

Open questions:
* Storage format. The easiest way is to save every URL into HTML file.
  Batching links into one file-database per root/favorite might be more efficient, especially for indexing.

* Reading of Firefox/Chrome export files (XML).
  Should we convert those files into our format or just adapt to them?

mdrlzy:
We also need to figure out how to generate a preview for the link.
- The file name can be replaced with the page title.
- Parsing an image. For different sites, the methods will be different (e.g. YouTube, Instagram)

mdrlzy:
Link preview(title, description, image) can be obtained using [ogp](https://ogp.me/).

kirillt:
Follow-up: https://github.com/ARK-Builders/ARK-Navigator/issues/235

Issue #33, opened 2021-08-25
Previews for video files


kirillt:
Some random frame from the video would be enough.
It can be useful to overlay also duration of the video and quality.

kirillt:
Necessary to extend `PreviewGenerators` object (#122).

PR #102, opened 2021-10-09
33 preview for video gif pdf

CloudLevi:


app/src/main/java/space/taran/arknavigator/utils/Tags.kt
```diff
@@ -20,7 +20,7 @@ class Converters {
             .split(',')
             .map { it.trim() }
             .filter {
-                if (it.isEmpty()) throw AssertionError("No tag can be empty")
+                if (it.isEmpty()) return@filter false
```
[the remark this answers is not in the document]
CloudLevi: I don't think so. As you can see, instead of throwing an exception, we return false now. That means, that filtering won't accept this file, because it doesn't reach a "return true" statement.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PlainResourcesIndex.kt
```diff
@@ -96,13 +99,47 @@ class PlainResourcesIndex internal constructor (
...
     }
 
+    private fun generatePdfPreviews() {
+        val previewsFolder = getPdfPreviewsFolder()
+        val savedPreviews = getSavedPdfPreviews()
+
+        metaByPath.forEach {
+            val path = it.key
+            var out: FileOutputStream? = null
+            val id = computeId(path)
```
mdrlzy, line 118: We already have an `id` in the meta.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -18,38 +32,181 @@ enum class Sorting {
...
+}
+
+const val PDF_PREVIEW_FOLDER_NAME = "pdf_preview"
+
+class PreviewFile(
+    val fileType: FileType? = null,
+    val file: Path? = null,
+    val fileExtension: String? = null,
+    val extraInfo: MutableMap<Preview.ExtraInfoTag, String>? = null
+) {
```
mdrlzy, line 50: What is the difference between a Preview and a PreviewFile?

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -18,38 +32,181 @@ enum class Sorting {
...
+                        fileExtension = extensionWithoutDot(file)
+                    )
+                else -> PreviewFile(fileExtension = extensionWithoutDot(file))
+            }
+        }
+
+        private fun getPdfPreview(file: Path): PreviewFile {
+            val id = computeId(file)
+            val savedPreviews = getSavedPdfPreviews()
+            return if (savedPreviews?.contains(id) == true) {
```
mdrlzy, line 83: Not sure here we need to recalculate the `id`. But with the current architecture it doesn't seem like it can be done otherwise.

app/src/main/java/space/taran/arknavigator/ui/fragments/utils/Preview.kt
```diff
@@ -8,17 +9,31 @@ enum class PredefinedIcon {
     FOLDER, FILE
 }
 
-data class Preview(val predefined: PredefinedIcon? = null, val image: Path? = null) {
+data class Preview(
+    val predefinedFolder: PredefinedIcon? = null,
```
mdrlzy, line 13: Maybe `predefinedIcon`?

mdrlzy (commented):
Maybe we should create a `PreviewRepo` that can be delivered via DI, can generate previews and provide them?

app/src/main/java/space/taran/arknavigator/ui/fragments/utils/Preview.kt
CloudLevi: > Maybe `predefinedIcon`?

Initially, "predefined" was used to determine if the file has a file or folder icon. With current implementation, this variable serves only one purpose - to determine whether the icon has a folder icon, or a variation of the file icon. Perhaps, in the future we will have some differentiations of the folder icon, hence the "PredefinedIcon" class. The name itself "predefined" was very confusing when writing if-else statements for icon sources, so I renamed it. I will turn it back.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
CloudLevi: This is a good point. It's possible to pass the id beforehand, but:
Out of the 3 places, where the method `Preview.provide(path: Path)` is called, the id is currently available in `GalleryPresenter` and `ResourcesGridPresenter`. However in `FoldersWalker` it's not directly available yet. I will remake it so that when we have the id available, we won't have to compute it.

mdrlzy:
> > Maybe we should create a `PreviewRepo` that can be delivered via DI, can generate previews and provide them?
> 
> To me, seems like overkill for this kinda task. Are there many more places where we'll implement previews?

If we are implementing "Link to Web" resource type, then I think we will need to show a preview before adding a resource on another screen.

We can also keep a cache for the preview paths in the application lifecycle instead of looking for previews in the file system every time.

CloudLevi:
Made PreviewRepo. Not sure how much of the logic is intented to be moved to the PreviewRepo. What are your thoughts? For example, function createPdfPreview - seems like it still should stay in FileUtils.

kirillt:
@mdrlzy
> If we are implementing "Link to Web" resource type, then I think we will need to show a preview before adding a resource on another screen.

Could you elaborate this?

app/src/main/java/space/taran/arknavigator/ui/fragments/utils/Preview.kt
kirillt: `PredefinedIcon` has 2 constructors: `FOLDER` and `FILE`.
`FILE` isn't used anymore, so just `val isFolder: Boolean` should be enough.
We don't need this enum anymore.

app/src/main/java/space/taran/arknavigator/ui/fragments/utils/Preview.kt
```diff
@@ -1,24 +1,28 @@
...
-            if (Files.isDirectory(path)) {
-                return Preview(predefined = PredefinedIcon.FOLDER)
-            }
-
-            val image = provideIconImage(path)
-                ?: return Preview(predefined = PredefinedIcon.FILE)
+data class Preview(
+    val predefinedIcon: PredefinedIcon? = null,
+    val previewPath: Path? = null,
+    val fileType: FileType? = FileType.UNDEFINED,
```
kirillt, line 16: 1. `fileType` should be named as `resourceType` since we might support non-single-file resources in future (like bookmarks #95).
2. Type of a resource should be taken from resources index as well as `extraInfo`, I'll push a commit with small modification and stubs to fill with concrete implementation.

app/src/main/java/space/taran/arknavigator/mvp/view/item/FileItemViewHolder.kt
```diff
@@ -19,11 +19,66 @@ class FileItemViewHolder(private val binding: ItemFileGridBinding) :
...
+            binding.iv.setImageResource(imageForPredefinedIcon(icon.predefinedIcon))
         } else {
-            loadImage(icon.image!!, binding.iv)
+            when (icon.fileType) {
+                FileType.GIF -> loadGifThumbnailWithPlaceHolder(
+                    icon.previewPath!!,
+                    imageForPredefinedExtension("gif"),
+                    binding.iv)
+                FileType.PDF -> {
+                    //Temporary workaround for asynchronous loading.
```
kirillt, line 31: What makes this code temporary? If we load the preview in lazy manner, then we should display progress spinner.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -18,38 +33,132 @@ enum class Sorting {
...
+        resolutionPair.containsAll(listOf(854, 480)) -> "480p"
+        resolutionPair.containsAll(listOf(1280, 720)) -> "720p"
+        resolutionPair.containsAll(listOf(1920, 1080)) -> "1080p"
+        resolutionPair.containsAll(listOf(2560, 1440)) -> "1440p"
+        resolutionPair.containsAll(listOf(3840, 2160)) -> "2160p"
+        else -> null
+    }
+}
+
+fun createPdfPreview(filePath: Path, context: Context? = null): Bitmap {
```
kirillt, line 140: Let's move this logic into dedicated place and not pollute `FileUtils` :)
This will also allow us to declare values like `pageNumber` as constants.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -18,38 +33,132 @@ enum class Sorting {
...
+}
+
+fun getSavedPdfPreviews(): List<Long?>? =
+    getPdfPreviewsFolder()
+        .listFiles()
+        ?.map {
+            it.nameWithoutExtension.toLongOrNull()
+        }
+
+fun getVideoInfo(filePath: Path): MutableMap<Preview.ExtraInfoTag, String> {
```
kirillt, line 91: The same as with `createPdfPreview` — let's create separate file for video parsing.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PreviewRepo.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    fun providePreview(path: Path, resourceID: ResourceId? = null): Preview {
+        if (Files.isDirectory(path)) {
+            return Preview(predefinedIcon = PredefinedIcon.FOLDER)
+        }
+
+        return createPair(path, resourceID)
+    }
+
+    private fun createPair(file: Path, resourceID: ResourceId?): Preview {
```
kirillt, line 21: I guess, name of this method left from some older code. Anyway, "Pair" is too ambiguous noun in the name.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -102,8 +211,16 @@ fun findLongestCommonPrefix(paths: List<Path>): Path {
...
-fun extension(path: Path): String =
-    path.extension
+fun getFileSizeMB(path: Path): Int =
+    (path.toFile().length() / (1024 * 1024)).toInt()
+
+fun extension(path: Path): String {
+    return ".${path.extension.lowercase()}"
+}
+
+fun extensionWithoutDot(path: Path): String {
```
kirillt, line 221: Can we use only 1 method of these 2? Do we really need extensions with dot?

kirillt:
I've added `getMeta()` method into `ResourcesIndex` and also modified `ResourceMeta` to contain type-specific data. Sorry, I had no time to implement type-specific data itself and put stubs (see `//TODO: PR #33` in the code). The idea is that we have to generate metadata like video duration/resolution during indexing and save it into `extra` field.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PreviewRepo.kt
```diff
@@ -0,0 +1,56 @@
...
+                Preview(
+                    fileType = FileType.GIF,
+                    previewPath = file,
+                    fileExtension = extensionWithoutDot(file)
+                )
+            else -> Preview(fileExtension = extensionWithoutDot(file))
+        }
+    }
+
+    private fun getPdfPreview(file: Path, resourceID: ResourceId? = null): Preview {
```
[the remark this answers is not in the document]
CloudLevi: Not sure how you want to use Path for this part. Here, we compare the id of the resource to the list of saved previews. When a preview is saved, it's saved using it's calculated id. In this part, we see if there is a file with a name that matches the ID. I understand that we need to remove our dependency on the ID, since it's unreliable. what should be the naming convention for previews, so we can save and retrieve them efficiently?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PreviewsRepo.kt
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+    }
+
+    fun loadPreview(
+        targetView: ImageView,
+        preview: Preview,
+        extraMeta: ResourceMetaExtra?,
+        centerCrop: Boolean = false
+    ) {
```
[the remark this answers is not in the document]
CloudLevi: In the case of `fun loadExtraMeta` I was able to remake the implementation so that it doesn't have to work with views.
In this case, I'm not sure how to do this. The whole point of moving this method to PreviewsRepo was to remove this heavy logic from adapters, that just have a job of displaying the data.

The issue here is that there are a lot of cases, and it's hard to encapsulate them all into some action. For example, for gifs we call methods `loadGif` or `loadGifThumbnailWithPlaceHolder` depending on the type of image. PDF's launch a coroutine in the context of the view to creade a bitmap of preview, and others may call `loadCroppedImageWithPlaceHolder` or `loadImageWithTransition` depending on other cases. How can we remake this, but keeping the business logic outside of adapters?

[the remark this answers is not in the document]
mdrlzy: Maybe create `PreviewHelper` or `PreviewLoader` in `utils`, that can resolve the preview and load it into the view?

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -18,38 +20,48 @@ enum class Sorting {
...
+    val extension = extension(filePath)
+    return acceptedImageExt.contains(extension)
+}
 
-//might be a temporary file
-fun providePreview(file: Path): Path? =
-    if (isImage(file)) {
-        file
+fun isVideo(filePath: Path?): Boolean {
+    return if (filePath?.toFile()?.exists() == true && filePath.fileSize() > 0) {
```
kirillt, line 41: We must not call method `isVideo` if the file doesn't exist. We should handle this somewhere else, irrelevant on file type.

kirillt: Also, don't use `java.io.File`, use `java.nio.Path`. All methods like `exists` are possible with newer `java.nio` API.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -18,38 +20,48 @@ enum class Sorting {
...
-    val extension = ".${extension(filePath)}"
-    return acceptedImageExt.contains(extension)
+fun isFormat(filePath: Path?, ext: String): Boolean {
+    return extension(filePath) == ext
 }
 
-fun getFileActionType(filePath: Path): FileActionType{
-    return when(".${extension(filePath)}"){
+fun getPdfPreviewsFolder(): File =
+    Paths.get("${App.instance.cacheDir}/$PDF_PREVIEW_FOLDER_NAME").toFile()
```
kirillt, line 54: Not necessary to convert to `java.io.File`, it's possible to use `java.nio.Path` almost everywhere.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -102,8 +114,16 @@ fun findLongestCommonPrefix(paths: List<Path>): Path {
     return tailrec(ROOT_PATH, paths).first
 }
 
-fun extension(path: Path): String =
-    path.extension
+fun deleteFile(filePath: Path?) {
+    val file = filePath?.toFile()
+    if (file?.exists() == true) {
+        file.delete()
```
kirillt, line 120: `java.nio` API has method `Files.delete`

app/src/main/java/space/taran/arknavigator/ui/fragments/GalleryFragment.kt
```diff
@@ -196,39 +209,44 @@ class GalleryFragment(
...
+                        openIntentChooser(currentPosition, Intent.ACTION_VIEW)
+                    }
                 }
-            }
-            FileActionType.OPEN_ONLY_DETACH_PROCESS -> {
-                openFileEditFab.makeGone()
-                openResourceChooserFab.makeVisibleAndSetOnClickListener{
-                    openIntentChooser(currentPosition, Intent.ACTION_VIEW, true)
+                FileActionType.OPEN_ONLY_DETACH_PROCESS -> {
+                    //todo remove after thorough testing
```
kirillt, line 231: We should not forget about this.

app/src/main/java/space/taran/arknavigator/ui/fragments/utils/Preview.kt
```diff
@@ -1,24 +1,15 @@
...
-
-            val image = provideIconImage(path)
-                ?: return Preview(predefined = PredefinedIcon.FILE)
-
-            return Preview(image = image)
-        }
+data class Preview(
+    val isFolder: Boolean? = null,
+    val filePath: Path? = null,
+    val fileExtension: String? = null,
```
kirillt, line 8: `fileExtension` is always piece of data stored in `filePath`, let's use method to access it, not field.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PlainResourcesIndex.kt
```diff
@@ -87,13 +102,12 @@ class PlainResourcesIndex internal constructor (
...
-
-            metaByPath[it] = ResourceMeta(
-                id = id,
-                modified = Files.getLastModifiedTime(it))
-            pathById[id] = it
+        toInsert.forEach { path ->
+            val meta = ResourceMeta.fromPath(path)
+            metaByPath[path] = meta
+            pathById[meta.id] = path
+            if (isFormat(path, "pdf"))
```
kirillt, line 109: This logic should not be here, just call `previewsRepo.generatePreview(...)` and dispatch on type there. If no preview needed for certain type, then just do nothing inside `generatePreview`.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PlainResourcesIndex.kt
```diff
@@ -72,13 +81,19 @@ class PlainResourcesIndex internal constructor (
     }
 
     internal suspend fun reindexRoot(diff: Difference) = withContext(Dispatchers.IO) {
+        val savedDeletedMetas = mutableMapOf<Path, ResourceMeta?>()
         diff.deleted.forEach {
             pathById[metaByPath[it]!!.id]
+            savedDeletedMetas[it] = metaByPath[it]
```
kirillt, line 87: Just delete preview here. No need to collect items and then do it in one go since all data resides in memory here.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PlainResourcesIndex.kt
kirillt: Also, this generation is called twice in some cases, since the same call is placed in `persistResources`, which is called from `reindexRoot`.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/ResourceMeta.kt
```diff
@@ -0,0 +1,76 @@
...
+            isVideo(filePath) -> ResourceType.VIDEO
+            isPDF(filePath) -> ResourceType.PDF
+            isFormat(filePath, "gif") -> ResourceType.GIF
+            else -> ResourceType.UNDEFINED
+        }
+    }
+
+    abstract fun appData(): MutableMap<Preview.ExtraInfoTag, String>
+
+    abstract fun roomData(): ResourceExtra //TODO: PR #33
```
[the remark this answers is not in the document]
kirillt: `ResourceMeta` and `ResourceMetaExtra` are related to indexing mechanism, although they are used by previews generating mechanism.

We can use plain variables, but then we also need to store kind of `ResourceMetaExtra` in other field and dispatch on it wherever we want to use `ResourceMetaExtra`. So I've created it as interface with methods to implement depending on kind. It's not the only way to implement it of course, but it's fine for the start.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/ResourceMeta.kt
```diff
@@ -0,0 +1,76 @@
...
+}
+
+enum class ResourceType {
+    IMAGE, VIDEO, GIF, PDF, UNDEFINED
+}
+
+
+abstract class ResourceMetaExtra {
+
+    abstract var filePath: Path?
```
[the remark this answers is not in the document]
kirillt: @mdrlzy The idea is that `ResourceMetaExtra` might be used for something else apart previews. E.g. we can sort resources by video quality or by number of pages in PDF.

@CloudLevi We must not store path inside of `ResourceMeta`. We already store metas by path in Room, so when we obtain `ResourceMeta`, we already have path, just pass it to methods which need it, don't duplicate data.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PreviewsRepo.kt
```diff
@@ -0,0 +1,198 @@
...
+    fun isZoomEnabled(preview: Preview, extraMeta: ResourceMetaExtra?) =
+        preview.isFolder == null &&
+                extraMeta?.type() != ResourceType.PDF &&
+                extraMeta?.type() != ResourceType.VIDEO
+
+    fun loadExtraMeta(
+        extraMeta: ResourceMetaExtra?,
+        resolutionTV: TextView? = null,
+        durationTV: TextView? = null
+    ) {
```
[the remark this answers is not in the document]
kirillt: I agree with @mdrlzy, `PreviewsRepo` should only know how to generate previews (picture + some flags). Views are should be managed in other places which should only ask the repo for previews and index for `ResourceMeta`.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PreviewsRepo.kt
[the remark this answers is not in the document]
kirillt: This method should be decomposed better. E.g. we don't need to generate previews during loading. We've already decided to generate them during indexing. After they are generated, it's only left to load them and this doesn't need the state at all. By the way, `PreviewsRepo` doesn't have any state (fields), all methods of it are static, we don't need it even to be `class`. Also, the name is confusing, "repo" usually is interface to some storage.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -18,38 +20,64 @@ enum class Sorting {
...
+    return extension(filePath) == "pdf"
+}
+
+fun isFormat(filePath: Path?, formatWithoutDot: String): Boolean {
+    return extension(filePath) == formatWithoutDot
 }
 
-fun getFileActionType(filePath: Path): FileActionType{
-    return when(".${extension(filePath)}"){
+fun getPdfPreviewsFolder(): File =
```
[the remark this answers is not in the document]
kirillt: Also, we don't need to store previews by kind of resource. Yes, right now we have previews only for PDFs, but when we'll have previews for other types, we'll store them in the same folder. So just `getPreviewsFolder`.

app/src/main/java/space/taran/arknavigator/ui/fragments/utils/Preview.kt
```diff
@@ -1,24 +1,15 @@
...
-                return Preview(predefined = PredefinedIcon.FOLDER)
-            }
-
-            val image = provideIconImage(path)
-                ?: return Preview(predefined = PredefinedIcon.FILE)
-
-            return Preview(image = image)
-        }
+data class Preview(
+    val isFolder: Boolean? = null,
```
kirillt, line 6: `Boolean?`, seriously? ;)

app/src/main/java/space/taran/arknavigator/mvp/model/repo/PreviewsRepo.kt
```diff
@@ -0,0 +1,188 @@
...
+            isPlayButtonVisible = isVideo(path)
+        )
+    }
+
+    private fun getPdfPreview(file: Path, id: ResourceId?): Preview {
+
+        val encodedID = id?.toString() ?: ""
+        val savedPreviews = getSavedPdfPreviews()
+
+        return if (savedPreviews?.contains(encodedID) == true) {
```
kirillt, line 42: `== true` !!!

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
```diff
@@ -28,12 +29,28 @@ class ResourcesGridPresenter(
...
 
+    @Inject
+    lateinit var previewsRepo: PreviewsRepo
+
     private var resources = listOf<ResourceId>()
+    set(value) {
+        field = value
+        resourcesMeta = value.map { index.getMeta(it) }
+    }
+    private var resourcesMeta = listOf<ResourceMeta?>()
```
kirillt, line 40: This is unnecessary duplication. `ResourceId` is part of `ResourceMeta`.
Also it never can be `null`.

kirillt:
Work moved to https://github.com/ARK-Builders/ARK-Navigator/pull/122

Issue #115, opened 2021-10-19
Replicated images cache (previews/thumbnails)


kirillt:
The app is generating previews/thumbnails and storing them in _cache_ folder.

Since the app is supposed to be used coupled with replication mechanism (like [Syncthing](https://syncthing.net/)) we can avoid re-generation of caches if we persist them. It might be stored in hidden folder. Only relative paths must be used in this case.

In case of previews/thumbnails, we have to ensure that they are generated universally for any device — it doesn't make sense to replicate cache if it works well only on certain device. This may change in future, since it is good to generate different previews for different screen sizes. When it'll change, we would like to have preference in settings menu (#77) for disabling replication of images cache to other devices.

PR #122, opened 2021-11-06
33 Previews and extra data for resources

kirillt:
`ResourceMetaExtra` substructure of `ResourceMeta` introduced. This field is used to store dynamically-determined data which depends on `kind` of `ResourceMeta` (like video, documents, images etc.)

Previews are implemented using global singleton object and their generation is triggered during indexing. This has to be changed in future (see #115). We store generated previews in cache folder so far (this will be changed upon implementation of persisted/replicated cache). For images we don't generate previews, but just use themselves. For any kind of resource, besides preview, thumbnail is also generated which is preview scaled down to 72x108 and stored in separate folder. Existence of preview/thumbnail for given resource is done prepending constant path to its `id`.

What needs to be finished prior to merge into main:
* Room structures for storing `ResourceMetaExtra` must be verified: `ResourceExtra` and `ResourceWithExtra`
* Previews and thumbnails generation should be verified as well
* Extra text snippets should be overlayed for videos and PDFs (see method `draw()` in `VideoMetaExtra` and `DocumentMetaExtra`)

Also would be good:
~~* Previews for videos (#33) is what gave name to this PR, but we could finish it later actually~~
~~* Make sortings in resources grid use `ResourceMeta` fields `size` and `modified` instead of querying file system~~
~~* Verify coroutines for loading thumbnails (see usage of `ThumbnailTarget`)~~


mdrlzy:
Glide sometimes loading wrong image on RecyclerView. 
Related issue: https://github.com/bumptech/glide/issues/2847
<p align="center">
 |

</p>

app/src/main/java/space/taran/arknavigator/mvp/model/repo/extra/DocumentMetaExtra.kt
```diff
@@ -0,0 +1,22 @@
...
+        setOf("pdf", "txt", "doc", "docx", "odt", "ods", "md")
+
+    fun extract(path: Path): ResourceMetaExtra? = null
+
+    fun draw(extra: ResourceMetaExtra, pagesTV: TextView, verbose: Boolean) {
+        val pages = extra.data[MetaExtraTag.PAGES]
+        if (pages != null) {
+            val label = if (verbose) "$pages pages" else "$pages"
+            pagesTV.textOrGone(label)
+        }
```
mdrlzy, line 20: Maybe the text generation can be left here, but the work with the view should be in the `ui` package
Same for `ImageMetaExtra` and `VideoMetaExtra`

app/src/main/java/space/taran/arknavigator/ui/fragments/preview/PreviewGenerators.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.graphics.Bitmap
+import space.taran.arknavigator.ui.fragments.preview.generator.PdfPreviewGenerator
+import java.nio.file.Path
+
+object PreviewGenerators {
+    var BY_EXT: Map<String, (Path) -> Bitmap> = mapOf(
+        "pdf" to { path: Path -> PdfPreviewGenerator.generate(path) }
+        //todo moar
+    )
+}
```
mdrlzy, line 12: I think preview generation should be in `model` package

app/src/main/java/space/taran/arknavigator/mvp/model/repo/extra/DocumentMetaExtra.kt
kirillt: Yes, you are right. Feel free to change this.

app/src/main/java/space/taran/arknavigator/ui/fragments/preview/PreviewGenerators.kt
kirillt: Agree :+1:

app/src/main/java/space/taran/arknavigator/ui/fragments/preview/PreviewAndThumbnail.kt
```diff
@@ -0,0 +1,149 @@
...
+                preview = previewPath(resource.id),
+                thumbnail = thumbnail)
+        }
+
+        fun forget(id: ResourceId) {
+            Files.deleteIfExists(previewPath(id))
+            Files.deleteIfExists(thumbnailPath(id))
+        }
+
+        @ExperimentalCoroutinesApi
```
kirillt, line 72: I guess we don't need it anymore

app/src/main/java/space/taran/arknavigator/ui/fragments/preview/PreviewAndThumbnail.kt
```diff
@@ -0,0 +1,149 @@
...
+        private const val THUMBNAIL_HEIGHT = 128
+        private const val COMPRESSION_QUALITY = 100
+
+        private fun previewPath(id: ResourceId): Path =
+            PREVIEWS_STORAGE.resolve(id.toString())
+        private fun thumbnailPath(id: ResourceId): Path =
+            THUMBNAILS_STORAGE.resolve(id.toString())
+
+        init {
+            if (PREVIEWS_STORAGE.notExists()) Files.createDirectories(PREVIEWS_STORAGE)
```
kirillt, line 45: Actually `Files.createDirectories` creates the directory if it doesn't exist.
That distinguishes this method from `Files.createDirectory` if I remember correctly.

PR #133, opened 2021-11-16
#115: Persisted/Replicated caches

kirillt:
Implementation of #115.

This branch is based on #102 — outdated version of #122.
Commits from that branch are not compatible with `main`.

If we decide to move forward with this PR, **we should put only the latest commit d44bdcaa845f233939c25487b879c7dece342232 on top of `main`**.


kirillt:
Closing this for now. The plan is to implement index/storage layer in separate library (Rust), which can be used on all other platforms (see #163). After that, #115 would be implemented directly in Rust since for replicated index we need to define format from scratch and we don't duplicate this work.

Issue #138, opened 2021-11-19
Thumbnails and previews missing


kirillt:
![_1](https://user-images.githubusercontent.com/581023/142668081-a23ac16a-1c05-43e4-a0f2-f59a5af79efe.jpg)
![_2](https://user-images.githubusercontent.com/581023/142668093-b666c3c8-017b-43ec-9507-e86955a8bc05.jpg)
![_3](https://user-images.githubusercontent.com/581023/142668099-488b529e-7c2c-4a00-8048-52295ee5cfe7.jpg)

kirillt:
Probably this happens for existing resources after the app restarted. New resources have their previews and thumbnails loaded.

This issue reproduces on my Android 9 device and it doesn't on Android 11.

kirillt:
The way to reproduce (I guess, on any device, but better to check):
1. Index a folder, previews and thumbnails must be visible.
2. Close the app.
3. Go to Android settings, find Apps section, locate the app.
4. Go to Storage section of the app, press "clear cache".
5. Run the app again, open the same folder.
Expected: previews and thumbnails are still visible.
Actual: previews and thumbnails are missing.

Explanation: apparently, we store the index as app's _data_ and previews/thumbnails as app's _cache_.
The index still contains all resources, so re-indexing doesn't heppen, but previews generation is triggered only during indexing.

vedranqa:
Creating folder: 
space.taran.arknavigator D/previews: previews provided in **12553ms**
space.taran.arknavigator D/resources-index: 1572 resources returned

After cleaning cache:
/space.taran.arknavigator D/previews: previews provided in **25169ms**
sace.taran.arknavigator D/resources-index: 1572 resources returned


Phone: Xiaomi A2
Android: 10
Folder: 1572 files, ~400mb size

Used build 165

kirillt:
@vedranqa that's strange.. no mistakes? Was the second result (after cleaning cache) _opening_ or _creating_ the folder?

vedranqa:
@kirillt this second case _After cleaning cache_ is **Opening** same folder like in first case.

kirillt:
Implemented in #143

vedranqa:
This is active on main build (id 204) after deleting cache we dont have thumbnails and previews.

vedranqa:
Build 204 issue is active.

Build 208 seems ok 


We can resolve this task. 
[...]

PR #143, opened 2021-11-20
#138: Thumbnails and previews missing

kirillt:
Before, we were generating previews only for added resources and assuming later that all indexed resources already have at least thumbnails (images don't have previews, only thumbnails).

Now, we check all existing resources for having previews/thumbnails generated.
If there is a thumbnail for a resource already, nothing happens.
If there is no such, new preview and thumbnails generated.

Also, time measurements with writing the results into log added.

**Help needed:** it is a good idea to handle exceptions coming from preview/thumbnail generation since they are failing whole coroutine and the rest of resources don't get their previews even if we could continue.
See `//todo: handle exceptions to not fail the rest of previews generation`

### DIFF
--- a/app/src/main/java/space/taran/arknavigator/utils/ImageUtils.kt
+++ b/app/src/main/java/space/taran/arknavigator/utils/ImageUtils.kt
@@ -1,6 +1,7 @@
 package space.taran.arknavigator.utils
 
 import android.graphics.drawable.Drawable
+import android.util.Log
 import android.widget.ImageView
 import com.bumptech.glide.Glide
 import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions
@@ -28,6 +29,7 @@ object ImageUtils {
     }
 
     fun loadZoomImageWithPlaceholder(image: Path?, placeHolder: Int, view: ImageView) {
+        Log.d(IMAGES, "loading image $image")
         view.setImageResource(placeHolder)
         view.autoDisposeScope.launch {
             withContext(Dispatchers.Main) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/space/taran/arknavigator/utils/ImageUtils.kt
```diff
@@ -54,7 +58,7 @@ object ImageUtils {
 
     fun loadImageWithPlaceholder(image: Path?, placeHolder: Int, view: ImageView) {
         Log.d(IMAGES, "loading image $image")
-        view.setImageResource(placeHolder)
+        //view.setImageResource(placeHolder)
```
kirillt, line 61: Commented-out code

app/src/main/java/space/taran/arknavigator/utils/ImageUtils.kt
```diff
@@ -65,4 +69,25 @@ object ImageUtils {
...
+
+    fun <T> glideExceptionListener() = object: RequestListener<T>{
+        override fun onLoadFailed(
+            e: GlideException?,
+            model: Any?,
+            target: Target<T>?,
+            isFirstResource: Boolean
+        ): Boolean {
+            Log.d(GLIDE, "load failed with message: ${e?.message} for target of type: ${target?.javaClass?.canonicalName}")
+            //Glide docs: "true if the listener has handled updating the target for the given exception."
```
kirillt, line 81: I think, this commentary is redundant, other developer should read official docs before trying to understand custom code :)

Issue #109, opened 2021-10-17
App crashed when returning to Resources Grid after resource deletion in Gallery screen


melvin4u445:
Steps:
1. Open Navigate data
2. Open an image 
3. Long-tap the delete button

```--------- beginning of crash
10-17 19:31:13.170 19569 19569 E AndroidRuntime: FATAL EXCEPTION: main
10-17 19:31:13.170 19569 19569 E AndroidRuntime: Process: space.taran.arknavigator, PID: 19569
10-17 19:31:13.170 19569 19569 E AndroidRuntime: java.lang.NullPointerException
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at space.taran.arknavigator.mvp.model.repo.PlainResourcesIndex.remove(PlainResourcesIndex.kt:65)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at space.taran.arknavigator.mvp.model.repo.AggregatedResourcesIndex.remove(AggregatedResourcesIndex.kt:20)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at space.taran.arknavigator.mvp.presenter.GalleryPresenter$deleteResource$1.invokeSuspend(GalleryPresenter.kt:46)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:873)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:99)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:216)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:7211)
10-17 19:31:13.170 19569 19569 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #149, opened 2021-11-29
#109: Gallery deletion crash

CloudLevi:
Needs further testing. Removed unnecessary loading, and the crash when deleting a resource and going back is fixed.

### DIFF
--- a/app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/AggregatedResourcesIndex.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/AggregatedResourcesIndex.kt
@@ -7,6 +7,8 @@ class AggregatedResourcesIndex(
     private val shards: Collection<PlainResourcesIndex>)
     : ResourcesIndex {
 
+    var pendingRemovalIDs = mutableListOf<ResourceId>()
+
     override fun listResources(prefix: Path?): Set<ResourceMeta> =
         shards.flatMap { it.listResources(prefix) }
             .toSet()
@@ -18,7 +20,11 @@ class AggregatedResourcesIndex(
         tryShards { it.tryGetMeta(id) }
 
     override fun remove(id: ResourceId): Path =
-        tryShards { it.tryRemove(id) }
+        tryShards {
+            val path = it.tryRemove(id)
+            if (path != null) pendingRemovalIDs.add(id)
+            path
+        }
 
     private fun <R>tryShards(f: (shard: PlainResourcesIndex) -> R?): R {
         shards.iterator()
--- a/app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
@@ -130,10 +130,9 @@ class ResourcesPresenter(
 
         presenterScope.launch {
             if (tagsEnabled) {
-                gridPresenter.resetResources(listResources())
-                updateSelection(tagsSelectorPresenter.selection)
+                gridPresenter.resetSelection()
             } else {
-                resetResources(listResources(untaggedOnly = true))
+                applyTaggedResources(listResources(untaggedOnly = true))
             }
         }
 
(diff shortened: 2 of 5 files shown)


app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
```diff
@@ -116,23 +116,23 @@ class ResourcesPresenter(
 
     fun onViewResume() {
         tagsSelectorPresenter.calculateTagsAndSelection()
-        if (!tagsEnabled) {
-            presenterScope.launch {
-                resetResources(listResources(untaggedOnly = true))
-            }
+        if (this::storage.isInitialized && this::index.isInitialized) {
+            val aggregatedIndex = (index as? AggregatedResourcesIndex)?: return
```
kirillt, line 120: 1) This is rude break of incapsulation principles.
2) This is not robust logic: when you have an interface you must assume the object to be any of possible implementations. Here, the bug is fixed for sharded index, but not for plain one.
3) Don't touch indexes please, this removal logic doesn't belong to them.

kirillt (changes requested):
Looks like dirty hack to me. Here, you should establish communication between UI elements, without involving lower-level components. Casting of object given by interface to particular implementation is really bad practice (read McConnell please, or some newer book about design and programming patterns, also about incapsulation). Even if we wanted to use index to notify all dependent objects about removals, that must be implemented more carefully (with storing removed ids in shards and aggregating them in sharded index), but that's redundant anyway.

mdrlzy (commented):
The main reason for the crash is asynchronous update of the selection. That's why RecyclerView binds elements from the old list.

As it seems to me, there are 2 solutions here:
- `index.getPath` should return a nullable result
- wrap `index.getPath` in try-catch

kirillt:
Might be possible to just wait till updating coroutine finished and bind elements with RecyclerView only afterwards.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/AggregatedResourcesIndex.kt
```diff
@@ -7,6 +7,8 @@ class AggregatedResourcesIndex(
     private val shards: Collection<PlainResourcesIndex>)
```
darkovski88, line 7: maybe add some kind of documentation so when other people read the code can know what this class is actually for?

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
darkovski88: I have one thing to add here - why do we have a presenter within a presenter? What is the reason behind this?

kirillt: @darkovski88 good question, maybe @mdrlzy can answer

app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/AggregatedResourcesIndex.kt
kirillt: @darkovski88 yeah, we need to document core classes.
Maybe this will be what we add:
```
`AggregatedResourceIndex` is useful for "aggregated" navigation mode — a mode in which we navigate
through multiple indexed folders (roots). For a single-root navigation `PlainResourceIndex` can be used, and
both `AggregatedResourceIndex` and `PlainResourceIndex` should be transparently interchangeable,
meaning that both implementations support the same methods and they should not be used
otherwise as by `ResourceIndex` interface. Any component using `ResourceIndex` should
function in the same fashion independent on implementation of the index,
and only capability to look into multiple roots is gained
by passing `AggregatedResourceIndex` into the component. 
```
I am not sure if it's too wordy or not, I hope it explains the purpose of this class.

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
mdrlzy: Multiple presenters are needed to separate business logic.
`gridPresenter` is responsible for which resources will be displayed, also handling the click on the resource and sorting.
`tagsSelectorPresenter` is responsible for which tags will be displayed, also handling a click on a tag and calculating a selection.
`ResourcesPresenter` is a screen layer presenter, it is responsible for initializing the index, tags storage, and handling toolbar menu click. It is also needed for communication between `gridPresenter` and `tagsSelectorPresenter`.

kirillt:
Substituted with https://github.com/ARK-Builders/ARK-Navigator/pull/172

PR #155, opened 2021-12-08
#109: Gallery deletion crash

mdrlzy:
#109

### DIFF
--- a/app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
@@ -1,10 +1,7 @@
 package space.taran.arknavigator.mvp.presenter.adapter
 
 import android.util.Log
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
+import kotlinx.coroutines.*
 import ru.terrakok.cicerone.Router
 import space.taran.arknavigator.mvp.model.UserPreferences
 import space.taran.arknavigator.mvp.model.repo.index.ResourceId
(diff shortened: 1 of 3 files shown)


app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
```diff
@@ -51,6 +53,9 @@ class ResourcesGridPresenter(
     fun getCount() = selection.size
 
     fun bindView(view: FileItemView) {
+        if (!resourcesReady)
```
darkovski88, line 56: is it a better idea to have this isResourceReady to be dependent on something else - not based on which method from the presenter is actually being called. For example to be dependant of some state(if something has finished or not)?

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
```diff
@@ -79,6 +84,7 @@ class ResourcesGridPresenter(
 
     suspend fun updateSelection(selection: Set<ResourceId>) = withContext(Dispatchers.Default) {
         this@ResourcesGridPresenter.selection = resources.filter { selection.contains(it.id) }
```
darkovski88, line 86: maybe it's a better idea to have this filtration of resources to be taken care from another class which will serve as a filter for this kind of things. Basically the job of the presenter is to do the logic(not to be done on the view). Also if we extract these kind of logics to separate classes we will have cleaner code, separation of concerns, and unit tests will be more easy to do.

kirillt: @darkovski88 actually we have such an entity, _tags selector_. Classes `TagsSelectorPresenter` and `TagsSelectorAdapter`. Maybe the architecture can be improved here.

darkovski88: @kirillt as i can see here the code and app looks ok to me. I have no negative comments on this PR except maybe take some action to the ones that i've already added

darkovski88 (approved):
Looks OK to me..

kirillt:
Substituted with https://github.com/ARK-Builders/ARK-Navigator/pull/172

Issue #77, opened 2021-09-21
Settings menu


kirillt:
The app needs new screen for various tweaks, all settings from that menu must be persisted.

kirillt:
Desired options in the settings screen:
* crash reporting over internet
* images cache replication
* images thumbnails generation
* reset all preferences

kirillt:
This issue might be referenced by other issues needing a preference.
But menu itself is implemented, so I'm closing it.

PR #135, opened 2021-11-17
#29: Restorable Fragments

mdrlzy:
Issues covered: #29 #4

PR #156, opened 2021-12-12
#77: Settings screen with persisted choices

CloudLevi:
Issue #77
- Implemented ACRA into the project
- Created Settings screen with persisted choices


app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
```diff
@@ -39,8 +48,77 @@ class UserPreferences @Inject constructor(val context: Context) {
...
     private object PreferencesKeys {
         val SORTING_PREFERENCE = intPreferencesKey("sorting_preference")
         val SORTING_ORDER = booleanPreferencesKey("sorting_preference_is_ascending")
+        val CRASH_REPORT_PREFERENCE = intPreferencesKey("crash_report_preference")
+        val IMG_CACHE_REPLICATION_PREF = intPreferencesKey("img_cache_replication_preference")
+        val INDEX_REPLICATION_PREF = intPreferencesKey("index_replication_preference")
+    }
+
+    enum class CrashReport {
+        NONE, SEND_AUTOMATICALLY, DONT_SEND
```
kirillt, line 94: What is `NONE`? Sending with confirmation?

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
```diff
@@ -39,8 +48,77 @@ class UserPreferences @Inject constructor(val context: Context) {
...
+    }
+
+    enum class ImgCacheReplication {
+        NONE, ENABLED, DISABLED
```
kirillt, line 98: `ImgCacheReplication` and `IndexReplication` look just binary to me.

app/src/main/java/space/taran/arknavigator/mvp/presenter/SettingsPresenter.kt
```diff
@@ -0,0 +1,74 @@
...
+        Log.d(SETTINGS_SCREEN, "first view attached in SettingsPresenter")
+        super.onFirstViewAttach()
+
+        viewState.init()
+        presenterScope.launch {
+            notifyAllPreferences()
+        }
+    }
+
+    fun onCrashPreferenceClick(crashReport: CrashReport) {
```
kirillt, line 32: I would name it `onCrashReportingClick` to be consistent with the rest of methods.

app/src/main/java/space/taran/arknavigator/ui/fragments/SettingsFragment.kt
```diff
@@ -0,0 +1,196 @@
...
+
+            crashInfo.setOnClickListener {
+                showInfoDialog(
+                    requireContext(),
+                    title = R.string.what_are_crash_reports_,
+                    descrText = R.string.crash_reports_explanation
+                )
+            }
+
+            //todo: insert meaningful explanation for images cache replication
```
kirillt, line 81: We need to get rid of other `//todo`s instead of putting new ones. If something should be done, we must use issue tracker for it. Particularly, about these 2 `//todo`s — we can just remove them.

app/src/main/java/space/taran/arknavigator/ui/fragments/SettingsFragment.kt
```diff
@@ -0,0 +1,196 @@
...
+                    descrText = R.string.explanation_of_this_feature
+                )
+            }
+
+            //todo: insert meaningful explanation for index replication
+            indexReplicationInfo.setOnClickListener {
+                showInfoDialog(
+                    requireContext(),
+                    title = R.string.what_is_index_replication_,
+                    descrText = R.string.explanation_of_this_feature
```
kirillt, line 95: By the way, we can present link to GitHub issue here :)

app/src/main/java/space/taran/arknavigator/ui/fragments/SettingsFragment.kt
```diff
@@ -0,0 +1,196 @@
...
+        ContextCompat.getColor(requireContext(), colorID)
+
+    private fun crashReportFromButton(buttonID: Int): CrashReport {
+        val reversed = crashReportMap.entries.associate { (k, v) -> v to k }
+        return reversed.getOrDefault(buttonID, CrashReport.NONE)
+    }
+
+    private fun buttonFromCrashReport(crashReport: CrashReport): Int {
+        return crashReportMap[crashReport]
+            ?: throw AssertionError("CrashReport must be of known type")
```
kirillt, line 152: Almost good. It won't hurt to check invariants in debug builds indeed. But `crashReportMap` has only 2 entries and `crashReport` can be 1 of 3 values, `NONE` is legal value and it will cause `AssertionError`. `AssertionError`s must be thrown only when we are sure they are never happen. This is double-check for situations when we assume something. It is helpful when we are wrong and some invariant doesn't hold.

app/src/main/java/space/taran/arknavigator/ui/fragments/SettingsFragment.kt
```diff
@@ -0,0 +1,196 @@
...
+        val reversed = indexReplicationMap.entries.associate { (k, v) -> v to k }
+        return reversed.getOrDefault(buttonID, IndexReplication.NONE)
+    }
+
+    private fun buttonFromIndexReplication(indexReplication: IndexReplication): Int {
+        return indexReplicationMap[indexReplication]
+            ?: throw AssertionError("IndexReplication must be of known type")
+    }
+
+    private val crashReportMap = mapOf(
```
kirillt, line 175: From the names of these maps, it's not clear what exactly we map.
Good name should involve semantics of keys or values, or both.
Here we only know to what component the map relates.

app/src/main/java/space/taran/arknavigator/ui/App.kt
```diff
@@ -23,4 +28,16 @@ class App: Application() {
...
     }
+
+    override fun attachBaseContext(base: Context?) {
+        super.attachBaseContext(base)
+
+        initAcra {
+            buildConfigClass = BuildConfig::class.java
+            reportFormat = StringFormat.JSON
+            toast {
+                text = "Sorry, the app crashed. Developers will be notified."
```
kirillt, line 39: This initialization takes around 2 seconds when a crash happens.
Let's initialize it only if we have crash reporting enabled.

app/src/main/java/space/taran/arknavigator/utils/DialogUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+import android.content.Context
+import android.view.View
+import androidx.annotation.StringRes
+import androidx.appcompat.app.AlertDialog
+import androidx.core.content.ContextCompat
+import space.taran.arknavigator.R
+import space.taran.arknavigator.databinding.DialogInfoBinding
+
+@SuppressLint("ClickableViewAccessibility")
+fun showInfoDialog(
```
kirillt, line 13: Awesome feature :+1:

kirillt (commented):
Good work, some polishing needed.

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
```diff
@@ -39,8 +48,78 @@ class UserPreferences @Inject constructor(val context: Context) {
...
+        NONE, SEND_AUTOMATICALLY, DONT_SEND
+    }
+
+    enum class ImgCacheReplication {
+        ENABLED, DISABLED
+    }
+
+    enum class IndexReplication {
+        ENABLED, DISABLED
+    }
```
mdrlzy, line 106: I think boolean will be enough

app/src/main/java/space/taran/arknavigator/utils/DialogUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+    @StringRes title: Int,
+    @StringRes descrText: Int,
+    @StringRes posButtonText: Int = R.string.ok,
+    @StringRes negButtonText: Int? = null,
+    posButtonCallback: (() -> Unit)? = null,
+    negButtonCallback: (() -> Unit)? = null
+) {
+    val builder = AlertDialog.Builder(context, R.style.AppTheme_AlertDialogStyle)
+    val view = View.inflate(context, R.layout.dialog_info, null)
+    val binding = DialogInfoBinding.bind(view)
```
mdrlzy, line 24: We should use the FragmentDialog and Fragment Result API. You can use the `EditTagsDialogFragment`(PR #135) as sample.

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
```diff
@@ -39,8 +48,38 @@ class UserPreferences @Inject constructor(val context: Context) {
...
     }
 
+    suspend fun setCrashReportEnabled(enabled: Boolean?) {
+        dataStore.edit { preferences ->
+            preferences[PreferencesKeys.CRASH_REPORT_PREFERENCE] = enabled?: BuildConfig.DEBUG
+        }
+    }
+
+    suspend fun isCrashReportEnabled(): Boolean =
+        dataStore.data.first()[PreferencesKeys.CRASH_REPORT_PREFERENCE]?: BuildConfig.DEBUG
```
kirillt, line 58: This seems to excessive since we already use it in `setCrashReportEnabled`

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
```diff
@@ -16,6 +17,14 @@ class UserPreferences @Inject constructor(val context: Context) {
 
     private val dataStore = context.preferencesDatastore
 
+    suspend fun clearPreferences() {
+        setSorting(Sorting.DEFAULT)
+        setSortingAscending(true)
+        setCrashReportEnabled(null)
+        setCacheReplicationEnabled(true)
```
kirillt, line 24: Let's set unimplemented features to `false`.
Can we also make these toggles inactive (gray and non-switchable)?
Also a toast notification by pressing on them would be nice.

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
```diff
@@ -39,8 +48,38 @@ class UserPreferences @Inject constructor(val context: Context) {
...
+        dataStore.data.first()[PreferencesKeys.CRASH_REPORT_PREFERENCE]?: BuildConfig.DEBUG
+
+    suspend fun setCacheReplicationEnabled(enabled: Boolean) {
+        dataStore.edit { preferences ->
+            preferences[PreferencesKeys.IMG_CACHE_REPLICATION_PREF] = enabled
+        }
+    }
+
+    suspend fun isCacheReplicationEnabled(): Boolean =
+        dataStore.data.first()[PreferencesKeys.IMG_CACHE_REPLICATION_PREF] ?: true
```
kirillt, line 67: Do we really need to duplicate default values in `clearPreferences` and `isSomethingEnabled` methods?
If yes, let's make constants and use them to avoid inconsistency.

app/src/main/java/space/taran/arknavigator/ui/fragments/FoldersFragment.kt
```diff
@@ -63,7 +63,7 @@ class FoldersFragment: MvpAppCompatFragment(), FoldersView, BackButtonListener {
 
     override fun init() {
         Log.d(FOLDERS_SCREEN, "initializing FoldersFragment")
-        (activity as MainActivity).setSelectedTab(0)
+        (activity as MainActivity).setSelectedTab(1)
```
kirillt, line 66: Do these numbers depend on order of initialization somewhere else?
If we can make them explicit that would be great.
If no, just define constants somewhere near to the place where they are determined.

app/src/main/java/space/taran/arknavigator/ui/fragments/SettingsFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    private fun showInfoDialog(
+        @StringRes titleRes: Int? = null,
+        @StringRes descriptionRes: Int? = null,
+        titleString: String? = null,
+        descriptionString: String? = null
+    ) {
+        val title = titleString ?: titleRes?.let { getString(it) }
+        ?: throw AssertionError("No value passed for title!")
```
kirillt, line 112: Good that you want to verify invariants, but here it looks a bit strange.
`titleString` is loaded earlier, so we can verify it there and make the argument non-nullable.
But we control state of these String variables, so it's redundant here.

`AssertionError` is useful when we have implicit invariants, some logical conditions we assume to be true.
Here the variable comes just from our condition, so it is explicit.

Also, it's not really critical to verify here, because user will not loose her data.

app/src/main/res/values/strings.xml
```diff
@@ -15,4 +15,14 @@
...
+    <string name="ok">OK</string>
+    <string name="yes">Yes</string>
+    <string name="no">No</string>
+    <string name="what_are_crash_reports_">What are crash reports?</string>
+    <string name="what_is_image_replication_">What is images cache replication?</string>
+    <string name="what_is_index_replication_">What is index replication?</string>
+    <string name="explanation_of_this_feature">Explanation of this feature</string>
+    <string name="are_you_sure_">Are you sure?</string>
+    <string name="all_preferences_will_be_reset_to_default">All user preferences will be reset to default. Do you wish to proceed?</string>
+    <string name="crash_reports_explanation">We are constantly working to improve ARK Navigator. However, you may still encounter issues while using the app. If you choose "Send automatically", we will automatically get notified if you run into a fatal crash of the app, so we can fix the issue.</string>
```
kirillt, line 27: :+1:

kirillt (commented):
Good work, but minor changes are still needed.

app/src/main/java/space/taran/arknavigator/ui/fragments/SettingsFragment.kt
kirillt: Btw, maybe some of `AssertionError`s placed by me should be normal exceptions in fact. We should check.

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
CloudLevi: This takes care of one scenario: On a fresh app, there is no saved state for `CRASH_REPORT_PREFERENCE`. When we select this option at least once or clear preferences - yes, it will be saved, and this nullability check won't be used.

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
CloudLevi: > Let's set unimplemented features to `false`. Can we also make these toggles inactive (gray and non-switchable)? Also a toast notification by pressing on them would be nice.

As I understand, `setCrashReportEnabled(null)` should still remain the same, so that the app can set this feature by default based on the flavor of the build?

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
CloudLevi: > Do we really need to duplicate default values in `clearPreferences` and `isSomethingEnabled` methods? If yes, let's make constants and use them to avoid inconsistency.

Yes, because we should take care of that one scenario on a freshly installed app. This scenario also extends whenever the user clears the caches of the app.

app/src/main/java/space/taran/arknavigator/mvp/model/UserPreferences.kt
kirillt: @CloudLevi yes, that's fine here (we will have crash reporting soon).

kirillt (approved):

Issue #25, opened 2021-08-17
Fix warnings


kirillt:
This issue will be updated with various debt about warnings.

**MainActivity.kt: (55, 27):** 'setOnNavigationItemSelectedListener(BottomNavigationView.OnNavigationItemSelectedListener?): Unit' is deprecated. Deprecated in Java

**MainActivity.kt: (168, 15):** 'onActivityResult(Int, Int, Intent?): Unit' is deprecated. Overrides deprecated member in 'androidx.activity.ComponentActivity'. Deprecated in Java

kirillt:
Warning: **The 'kotlin-android-extensions' Gradle plugin is deprecated.** Please use this migration guide (https://goo.gle/kotlin-android-extensions-deprecation) to start working with View Binding (https://developer.android.com/topic/libraries/view-binding) and the 'kotlin-parcelize' plugin.

kirillt:
Also, linter results should be fixed. See `lint-results` artifact in our builds: https://github.com/kirillt/ark-navigator/actions/runs/1140276919

PR #161, opened 2021-12-16
#25: Fix warnings

CloudLevi:


CloudLevi:
Currently started reworking the permission request. There is a bug here, but in a very rare scenario. Steps to reproduce:
1. On a fresh app, open the app
2. Grant storage permissions
3. Add a root
4. Without closing the app completely, go to app settings on the device, and revoke the storage permissions
5. Go back into the app (the app should ask for storage permissions again)
6. Deny the permission (saved root folder should still be present)
7. Click on those roots and click the navigate button, the app may crash or behave unexpectedly

Otherwise, I tried blocking all interactions until we have storage permissions. How do you think we should handle that event, and similar ones? My suggestions are:
1. Clear the fragments, so the user will end up with a blank screen and nothing to interact with
2. Have some kind of a custom Router, that will check storage permissions when trying to make Fragment transactions
3. (I think it's a bad one, as it can cause memory leaks, but will toss it here) store "hasStoragePermissions" in MainActivity or other place as a variable, and check it from the Presenter and other places thoughout the app

kirillt:
@CloudLevi could you register this bug as issue?

Also, what does this warning mean? I don't see any usage of `ANDROID_NDK_HOME` in build scripts.
> WARNING: Support for ANDROID_NDK_HOME is deprecated and will be removed in the future. Use android.ndkVersion in build.gradle instead.

Also please pull commits from `main`.

kirillt:
Btw, linter doesn't upload results last month. Could you also look what happened to it?

Issue #159, opened 2021-12-15
Crash during start-up: nonexistent app data folders


kirillt:
Fresh installation on Samsung Galaxy J8 (Android 9) resulted in the following error:
```
12-15 17:25:10.054   442   442 E AndroidRuntime: Process: space.taran.arknavigator, PID: 442
12-15 17:25:10.054   442   442 E AndroidRuntime: java.nio.file.FileSystemException: /dev/null/Android/data/space.taran.arknavigator/files: Not a directory
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at sun.nio.fs.UnixPath.toRealPath(UnixPath.java:837)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at space.taran.arknavigator.utils.FileUtilsKt.listDevices(FileUtils.kt:30)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at space.taran.arknavigator.mvp.presenter.FoldersPresenter$onFirstViewAttach$1.invokeSuspend(FoldersPresenter.kt:65)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:873)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:99)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:214)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:7078)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
12-15 17:25:10.054   442   442 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:974)
```

[...]

PR #162, opened 2021-12-17
#159: List devices filter

mdrlzy:
#159


kirillt:
Will push similar solution soon.

Issue #169, opened 2021-12-24
Performance of "quick tags" degraded


kirillt:
In 9d0b39d0c71bc495dc5d6ae960b56d9f43233e1f loading time of `EditTagsDialogPresenter` was mush faster:
```
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 343 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 173 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 142 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 113 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 121 milliseconds
---
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 222 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 85 milliseconds
```
Since eb128d30f7606c817df02d8cd8178f45be30cc65 loading time (body of the coroutine) is slower:
```
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 405 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 415 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 358 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 277 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 251 milliseconds
---
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 662 milliseconds
D/EditTagsDialogPresenter: `onFirstViewAttach()` took 277 milliseconds
```

Probably, `resolveRootAndFav` is the culprit (it loads folders from Room).

mdrlzy:
Fixed by #170

PR #170, opened 2021-12-25
#169 Quick tags perfomance

mdrlzy:
#169

### DIFF
--- a/app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourcesIndexRepo.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourcesIndexRepo.kt
@@ -62,8 +62,9 @@ class ResourcesIndexRepo(
     }
 
     suspend fun provide(rootAndFav: RootAndFav): ResourcesIndex = withContext(Dispatchers.IO) {
+        val roots = foldersRepo.resolveRoots(rootAndFav)
+
         provideMutex.withLock {
-            val roots = foldersRepo.resolveRoots(rootAndFav)
             val indexShards = roots.map { root ->
                 indexByRoot[root] ?: let {
                     val index = loadFromDatabase(root)
--- a/app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourcesIndexRepo.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourcesIndexRepo.kt
@@ -2,6 +2,8 @@ package space.taran.arknavigator.mvp.model.repo.index
 
 import android.util.Log
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 import space.taran.arknavigator.mvp.model.dao.ResourceDao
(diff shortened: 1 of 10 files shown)


app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourcesIndexRepo.kt
```diff
@@ -59,31 +62,28 @@ class ResourcesIndexRepo(
...
-        val indexShards = roots.map { root ->
-            indexByRoot[root] ?: let {
-                val index = loadFromDatabase(root)
-                indexByRoot[root] = index
-                index
+        provideMutex.withLock {
+            val roots = foldersRepo.resolveRoots(rootAndFav)
+            val indexShards = roots.map { root ->
+                indexByRoot[root] ?: let {
+                    val index = loadFromDatabase(root)
```
kirillt, line 69: I think, we can use `foldersRepo.resolveRoots(..)` without locking.
It has lock inside and we don't need locking for local variable `roots`.

About `indexByRoot` map: if we can use "read-write lock" instead of mutex, then we would do read-lock when we iterate its entries (no real locking if there is not write), and only do write-lock when we have no index loaded in particular entry. This way, in case we have all roots loaded, we don't write anything new into `indexByRoot`, so we do only read-lock which is much cheaper.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/tags/TagsStorageRepo.kt
```diff
@@ -11,29 +13,31 @@ class TagsStorageRepo(
...
     private val indexRepo: ResourcesIndexRepo
 ) {
+    private val provideMutex = Mutex()
     private val storageByRoot = mutableMapOf<Path, PlainTagsStorage>()
 
     suspend fun provide(rootAndFav: RootAndFav): TagsStorage = withContext(Dispatchers.IO) {
-        val allRoots = foldersRepo.query().succeeded.keys
-        val roots = if (rootAndFav.isAllRoots()) allRoots else listOf(rootAndFav.root!!)
+        provideMutex.withLock {
+            val roots = foldersRepo.resolveRoots(rootAndFav)
```
kirillt, line 21: We can do locking after resolving roots, the same as in `ResourcesIndexRepo.provide()`.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/tags/TagsStorageRepo.kt
```diff
@@ -11,29 +13,31 @@ class TagsStorageRepo(
...
-            } else {
-                val fresh = PlainTagsStorage(root, resources)
-                fresh.init()
-                fresh.cleanup(resources)
-                storageByRoot[root] = fresh
-                fresh
+            val storageShards = roots.map { root ->
+                val index = indexRepo.provide(root)
+                val resources = index.listAllIds()
+                if (storageByRoot[root] != null) {
```
kirillt, line 26: Similar to `ResourcesIndexRepo.provide()`, "read-write lock" would allow us to really lock only when we actually write something into the map.

kirillt (commented):
Looks good, but if we can adopt "read-write lock" then it should give much better performance (measurements won't hurt).

kirillt (approved):

Issue #164, opened 2021-12-21
Glitch with preview and thumbnails


vedranqa:
Opening files there is flash of preview icon before opening photo and the sam e on thumbnails. 
Details in video

https://user-images.githubusercontent.com/60543054/146911371-8f36dc55-96c2-4698-b0de-40a80fb6f5e8.mp4

Android 10
Xiaomi A2
Build 208

kirillt:
Fixed in #172

PR #172, opened 2021-12-27
#164: Grid glitch on return

mdrlzy:
Issues covered: #164 #109

### DIFF
--- a/app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
@@ -1,6 +1,7 @@
 package space.taran.arknavigator.mvp.presenter
 
 import android.util.Log
+import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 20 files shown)


kirillt:
Looks good, but here is issue with next preview not loading sometimes after deletion.

https://user-images.githubusercontent.com/581023/147541816-acf94ca7-5414-4796-a17f-eea824b86075.mp4

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
```diff
@@ -119,22 +120,17 @@ class GalleryPresenter(
         viewState.editResource(index.getPath(currentResource.id))
     }
 
-    fun onRemoveFabClick() {
+    fun onRemoveFabClick() = presenterScope.launch {
```
kirillt, line 123: Should we add `NonCancellable` here in order to prevent half-deletion?

kirillt (approved):

Issue #154, opened 2021-12-07
Automate code formatting upon every git commit


kirillt:
We should use this: https://github.com/pinterest/ktlint
1. Git Hooks must be used to run `ktlint` upon every commit (https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)
2. CI (Github Actions) must fail if code isn't formatted.
3. We should decide together about coding style and encode it in `ktlint` config.

kirillt:
Finalized in 9ed454267685d30423f6bb6bb240688a4a6a073d

PR #181, opened 2022-01-09
#154: ktlint implementation

kirillt:
@CloudLevi 
* Several Gradle versions seem to be mentioned (7.0.2 and 7.0.4)
* File `.editorconfig` should be used to specify rules of formatting
* Is pre-commit hook really working?
(I've tried it on Linux and it looks possible to commit non-nformatted code without problem)


CloudLevi:
@kirillt about gradle versions:
In `build.gradle`, we define the version for android gradle plugin.
In `gradle-wrapper.properties`, we define the gradle distribution version, which are different things. The `gradle-wrapper.properties` file was changed automatically upon updating the gralde version, so it should be the recommended supported verion of distribution for our gradle version, which is `7.0.4`

kirillt:
(Rebased on top of `main`)

kirillt:
Optimized pre-commit hooks (`ktlint` is run only if there are changed `.kt` files).
Optimized GitHub Actions workflows.

Issue #200, opened 2022-02-01
Sorting should be case insensitive


melvin4u445:
Files starting with a capital letter are shown first and files starting with a small letter are shown after that. 
<img src="https://user-images.githubusercontent.com/17034768/152032697-2f0e13e8-2692-45d1-b4c5-f30914b7ce33.jpg" width=40%> <img src="https://user-images.githubusercontent.com/17034768/152032716-db2d5aaa-bdb7-4a36-9120-8d671530eb15.jpg" width=40%>

kirillt:
I think it's difficult to determine what is desired regarding order of special characters..
Maybe we should just eliminate sorting by name sometimes :)

kirillt:
@sisco0 sorry for late response, I'm still working on my GitHub habits.. Thank you, your commit looks perfectly fine, don't hesitate to submit pull-request. Actually, I would notice PR quicker than mentioning :) Also don't hesitate to reach me in Telegram (@kirill_taran) or by e-mail (kirill@taran.space) for discussing paid tasks.

PR #219, opened 2022-02-25
Case insensitive sorting

sisco0:
Solves #200.

A new comparator behavior has been integrated so we could sort in a case insensitive order.


kirillt:
@sisco0 thank you! Could you also enable GitHub Actions in your fork, please? This way our QA would be able to download corresponding build.

kirillt (approved):
Thank you @sisco0 !

PR #210, opened 2022-02-12
#195,#196: virtual "kind" tags and jumping to Resource Grid

Kashif2016:
Hi Kirill, please check the pull request for these two tasks mentioned below.

https://github.com/ARK-Builders/ARK-Navigator/issues/195
https://github.com/ARK-Builders/ARK-Navigator/issues/196

Issue #196, opened 2022-01-29
Jumping from Gallery to Resource Grid with new selection


kirillt:
While walking through sequence of resources in Gallery mode, user's attention might get drawn by some specific tag on some resource. This tag is not necessarily part of initial user's query. The app must allow the user to see all other resources labeled with that tag by jumping to resource grid screen with new selection which would consist of the only clicked tag.
![feature](https://user-images.githubusercontent.com/581023/151675872-79be2527-6d22-4c21-a333-14248902be29.png)

Handling of **[back]** button is important here. We should maintain stack of screens. After the "jump", the user can go both directions: either to open gallery again from new selection, or to go back to gallery screen with old selection.

Issue #195, opened 2022-01-29
Virtual tags in query selector: resource "kinds"


kirillt:
Sometimes, it would be convenient to filter resources not only by tags but also by their type. In our app, we replace _file_ notion with _resource_ and _type_ with _kind_. Kinds are more generic: e.g. both `JPG` and `PNG` are _images_ and a user rarely cares about exact types of images. Usage of _kind_ notion also hints that, technically, it might be determined not only by file extension.

Tag selector in resource grid should provide user with auto-generated _virtual_ tags. For the beginning, these virtual tags must correspond to resource kinds. Virtual tags must be highlighted differently. They are different from normal tags in that it's not possible to label a resource with them manually, and they are not stored. They are similar to tags that every tag can filter selection and can be composed with other tags. This implies that virtual tags should also be sorted the same way as other tags in the selector.

From UI perspective, button in tag selector also should be created to turn on/off both generation and displaying of virtual tags.

Virtual tags must be also provided for _unlabeled selection_, i.e. when no tags are selected. If the user didn't turn virtual tags off, of course.

kirillt:
![1](https://user-images.githubusercontent.com/581023/153752800-50fa7856-6219-4b0a-a4ba-bd17c89733f2.jpg)

![2](https://user-images.githubusercontent.com/581023/153752810-c431c714-6ac0-4113-9f6c-e9117c5a6e24.jpg)

PR #220, opened 2022-03-13
 #195,#196: virtual "kind" tags and jumping to Resource Grid

kirillt:
This is follow-up PR based on previous work in #210
Comments from #210 must be resolved in this PR.

* Jumping from Gallery mode to new Resource grid containing clicked tag is implemented.
  But the code should work with new query in generic way — there is no sense to create special variable containing only 1 selected tag, also we already have `includedTags` variable in `TagSelector`, it can be re-used.

* If user wants to go **[back]**, the app must return to previous selection (right now, **[back]** results in empty selection).

* Also list of active folders is reset after **[back]** event (if we started in e.g. Pictures folder, when we are returning to previous selection, suddenly aggregated mode with all folders is triggered).

* UI of "kind" tags should be enhanced: `"Show Kind Tags"` label takes too much of space, we should invent special icon instead of toggle button. Style of tags in the selector also should be returned to initial one ("kind" tags should be still highlighted though).


app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/tagsselector/TagsSelectorPresenter.kt
```diff
@@ -3,95 +3,125 @@ package space.taran.arknavigator.mvp.presenter.adapter.tagsselector
...
+}
+
 class TagsSelectorPresenter(
     private val viewState: ResourcesView,
     private val prefix: Path?,
     private val scope: CoroutineScope,
     private val onSelectionChangeListener: KSuspendFunction1<Set<ResourceId>, Unit>
 ) {
+    @Inject
+    lateinit var stringProvider: StringProvider
```
kirillt, line 33: Is it possible to inject just `Context`? So it would be possible to remove code duplicate below (initializing `kindToString`). We have the same code in `TagsSelectorAdapter` which might be forgotten to be updated when we add more resource kinds. Ideally, this map creation should belong to `ResourceMeta` layer.

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/tagsselector/TagsSelectorPresenter.kt
```diff
@@ -261,51 +295,96 @@ class TagsSelectorPresenter(
...
+                        kindToString[it.kind]!!.startsWith(filter, false)
+                    }
+                }
+            }
 
-        unavailableTagsForDisplay = unavailableTags
-            .filter { it.startsWith(filter, false) }
+        unavailableTagsForDisplay = unavailableTagItems
+            .filter {
+                when (it) {
```
kirillt, line 312: Code duplicate. Maybe we should implement something like `KindTagItem extends Stringlike` where `Stringlike { fun gegtString(): String { ... } }`?

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/tagsselector/TagsSelectorPresenter.kt
```diff
@@ -261,51 +295,96 @@ class TagsSelectorPresenter(
...
+                    is TagItem.DefaultTagItem -> {
+                        it.tag.startsWith(filter, false)
+                    }
+                    is TagItem.KindTagItem -> {
+                        kindToString[it.kind]!!.startsWith(filter, false)
```
kirillt, line 317: I haven't anticipated this feature, looks nice :+1:

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/tagsselector/TagsSelectorPresenter.kt
mdrlzy: Extracted `kindToString` to `StringProvider`. We do not want to inject the `Сontext` into the business logic, because then it will be impossible to use unit tests, since we will depend on the android. It is better to have pure Kotlin and use abstractions that we can replace in tests.

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/tagsselector/TagsSelectorPresenter.kt
mdrlzy: Replaced with `filterTagItem` function

PR #223, opened 2022-03-16
#95: "Link to Web" resource kind 

mdrlzy:
#95


app/build.gradle
```diff
@@ -75,6 +75,8 @@ dependencies {
 
     implementation "com.google.android.material:material:1.4.0"
 
+    implementation 'com.beust:klaxon:5.5'
```
darkovski88, line 78: Im just curious why do we need this klaxon library?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/extra/LinkMetaExtra.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.beust.klaxon.Klaxon
+import space.taran.arknavigator.mvp.model.repo.index.MetaExtraTag
+import space.taran.arknavigator.mvp.model.repo.index.ResourceMetaExtra
+import java.nio.file.Path
+import java.util.zip.ZipFile
+
+object LinkMetaExtra {
+    val ACCEPTED_EXTENSIONS = setOf("link")
+    private const val JSON_FILE = "link.json"
+    val klaxon = Klaxon()
```
darkovski88, line 12: klaxon can be private

app/src/main/java/space/taran/arknavigator/ui/extra/DocumentExtraLoader.kt
```diff
@@ -7,11 +7,11 @@ import space.taran.arknavigator.utils.extensions.textOrGone
 
 object DocumentExtraLoader {
     fun load(extra: ResourceMetaExtra, pagesTV: TextView, verbose: Boolean) {
-        val pages = extra.data[MetaExtraTag.PAGES]
+        val pages = extra.data[MetaExtraTag.PAGES]?.toInt()
```
darkovski88, line 10: `toIntOrNull` is a better approach here in my opinion

app/src/main/java/space/taran/arknavigator/ui/extra/VideoExtraLoader.kt
```diff
@@ -23,20 +23,20 @@ object VideoExtraLoader {
             )
         }
 
-        val duration = extra.data[MetaExtraTag.DURATION]
+        val duration = extra.data[MetaExtraTag.DURATION]?.toInt()
```
darkovski88, line 26: `toIntOrNull()` instead of `toInt()` ?

app/src/main/java/space/taran/arknavigator/ui/fragments/GalleryFragment.kt
```diff
@@ -314,10 +330,11 @@ class GalleryFragment : MvpAppCompatFragment(), GalleryView, NotifiableView {
 
     private fun setupOpenEditFABs(kind: ResourceKind?) {
```
darkovski88, line 331: maybe it's a good time to start adding some documentation to the method definition. Maybe not all of the methods but methods which we thing are not good enough named for other devs. `setupOpenEditFABs`  doesn't really describe the method just right in my opinion. What does `FAB` mean for example?

app/build.gradle
mdrlzy: To work with json, do you mean replace it with kotlinx.serialization?

app/src/main/java/space/taran/arknavigator/ui/fragments/GalleryFragment.kt
mdrlzy: Maybe we should create a separate issue for this, but it seems to me that there will still be refactoring.
FAB stands for [floating action button](https://material.io/components/buttons-floating-action-button/android)

cnzqa:
can we fix this text under share button ?
image

kirillt:
@cnzqa what is the extension of this file? Did you add it from ARK Shelf?

mdrlzy:
@cnzqa Of course this needs to be fixed. Yes, we should add an icon for the .link files because previews can't always be generated.

@kirillt This is a .link file, just the preview was not generated.

cnzqa:
> @cnzqa what is the extension of this file? Did you add it from ARK Shelf?

Yes, I've used ARK Shelf

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/DocumentKindFactory.kt
```diff
@@ -1,22 +1,19 @@
...
-    val ACCEPTED_EXTENSIONS: Set<String> =
+object DocumentKindFactory : ResourceKindFactory {
+    override val acceptedExtensions: Set<String> =
         setOf("pdf", "txt", "doc", "docx", "odt", "ods", "md")
 
-    fun extract(path: Path): ResourceMetaExtra? {
-        val result = mutableMapOf<MetaExtraTag, String>()
-        if (extension(path) != "pdf") return null
+    override fun extract(path: Path): ResourceKind {
+        if (extension(path) != "pdf") return ResourceKind.Document()
```
kirillt, line 16: It was `null` before, why `ResourceKind.Document()` now?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/ResourceKind.kt
```diff
@@ -0,0 +1,79 @@
+package space.taran.arknavigator.mvp.model.repo.kind
+
+import space.taran.arknavigator.mvp.model.dao.ResourceExtra
+import space.taran.arknavigator.utils.extension
+import java.nio.file.Path
+import kotlin.reflect.full.isSubclassOf
+
+sealed class ResourceKind {
```
kirillt, line 8: Now it became a bit confusing name, but maybe it's still better than `ResourceMetaExtraImpl`...

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
```diff
@@ -105,8 +104,9 @@ class GalleryPresenter(
 
     fun onOpenFabClick() {
         Log.d(GALLERY_SCREEN, "[open_resource] clicked at position $currentPos")
-        if (currentResource.kind == ResourceKind.LINK) {
-            val url = currentResource.extra?.data?.get(MetaExtraTag.URL) ?: return
+        val kind = currentResource.kind
```
kirillt, line 107: Here the refactoring looks nice.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/ResourceKind.kt
```diff
@@ -0,0 +1,79 @@
...
+    class Document(val pages: Int? = null) : ResourceKind()
+
+    class Link(
+        val title: String? = null,
+        val description: String? = null,
+        val url: String? = null
+    ) : ResourceKind()
+
+    companion object {
+        private val sealedSubClasses = ResourceKind::class.sealedSubclasses.filter {
```
kirillt, line 28: Is it possible only with _sealed_ subclasses?

I would get rid of reflection at all and use enums again :) Reflection in general is slower and unreliable. Although we don't need speed when we extract subclasses, what I don't like in this file is that for every new kind developer would need to put declaration of it here. I would just make field-list containing all necessary implementations, and put everything related to particular kinds into separate files.

Also if it is possible to hide stuff like
```
Video::class ->
                    Video(
                        data[MetaExtraTag.HEIGHT]?.toLong(),
                        data[MetaExtraTag.WIDTH]?.toLong(),
                        data[MetaExtraTag.DURATION]?.toLong()
                    )
```
into kind-specific file, it would be also good.

Otherwise we risk overbloat this file when we add 10 or more kinds. Also, adding new kind, ideally, should be copying existing file and modifying it.

Refactoring looks nice in the rest of places.

app/src/main/java/space/taran/arknavigator/mvp/model/dao/ResourceExtra.kt
```diff
@@ -26,14 +27,84 @@ data class ResourceExtra(
...
+            is ResourceKind.Video -> mapVideo(id, kind)
+            is ResourceKind.Link -> mapLink(id, kind)
+            else -> emptyList()
+        }
+
+        private fun mapDocument(
+            id: ResourceId,
+            document: ResourceKind.Document
+        ): List<ResourceExtra> {
+            val extras = mutableListOf<ResourceExtra>()
```
kirillt, line 43: I think mutable list can be replaced by normal `map`.

app/src/main/java/space/taran/arknavigator/mvp/model/dao/ResourceExtra.kt
```diff
@@ -26,14 +27,84 @@ data class ResourceExtra(
...
+            return extras
+        }
+
+        private fun mapVideo(
+            id: ResourceId,
+            video: ResourceKind.Video
+        ): List<ResourceExtra> {
+            val valueMap = mutableMapOf<Int, String>()
+            video.height?.let {
+                valueMap[MetaExtraTag.HEIGHT.ordinal] = it.toString()
```
kirillt, line 62: Why not just generate `key to value` pairs in immutable map and map them into `ResourceExtra`s?

app/src/main/java/space/taran/arknavigator/mvp/model/dao/ResourceExtra.kt
```diff
@@ -26,14 +27,84 @@ data class ResourceExtra(
...
-            List<ResourceExtra> =
-            extra?.data?.entries?.map { (tag, value) ->
-                ResourceExtra(
-                    resource = id,
-                    ordinal = tag.ordinal,
-                    value = value
+        fun fromMetaExtra(id: ResourceId, kind: ResourceKind?):
+            List<ResourceExtra> = when (kind) {
+            is ResourceKind.Image -> emptyList()
+            is ResourceKind.Document -> mapDocument(id, kind)
```
kirillt, line 33: These methods should be placed into kind-specific modules.

app/src/main/java/space/taran/arknavigator/ui/extra/VideoExtraLoader.kt
kirillt: @darkovski88 this seems to not depend on user input, if we can't parse `Int` here, then we made error during saving.

app/src/main/java/space/taran/arknavigator/ui/fragments/GalleryFragment.kt
kirillt: I would name this action in more generic way, like `setupResourceActions()` or `setupResourceExternalOperations()` or `displayControls()`.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/DocumentKindFactory.kt
mdrlzy: Because before that we returned a nullable list of extras, now in any case we need to return a kind, and inside properties can be null.

kirillt (approved):
Sharing doesn't work but the rest is cool.
Follow-up issue: https://github.com/ARK-Builders/ARK-Navigator/issues/235

PR #237, opened 2022-04-19
Something is changed certain way to fix #123

### LINKED ISSUE #123: Tags editing dialog: [backspace] must delete last tag, [back] should discard any deletions, commit changes only by [done] event
1. When a user presses [back] button, the current input is saved. This is fine when we have unfinished input, because it is lost (as expected). But, if we deleted several tags and press [back] button — we don't undo deletion of tags.

2. Right now, we handle [done] as saving current state into tags storage. After [done] event we leave input field active. We should close the dialog after [done].

3. If we enter new tag and trigger [done], it becomes "chip". If we press [backspace] this new chip must be deleted. Chips should only be view of text input.

`EditTagsDialogPresenter` should store raw input and `EditTagsDialogFragment` only draw this input as chips.
Right now, the input stored only in the fragment. We also should store initial state in the presenter. We should not spawn coroutines and write to the storage before [done] is triggered. If [back] is triggered, we should discard all changes (this means no writes to storage was done during work with the dialog).

kirillt:
THIS IS DEMO PR SERVING AS EXAMPLE FOR FUTURE CONTRIBUTORS

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+THIS IS MY AWESOME CHANGE!
+
+
 # **ARK Navigator** ![example workflow](https://github.com/kirillt/ark-navigator/actions/workflows/build.yml/badge.svg)
 
 _ARK Navigator_ is the next-generation file browser built on the idea of tags.


README.md
```diff
@@ -1,3 +1,6 @@
+THIS IS MY AWESOME CHANGE!
```
kirillt, line 1: Reviewers will leave comments to your Pull Request.
Please fix all issues if there are any.

README.md
```diff
@@ -1,3 +1,6 @@
...
+
+
```
kirillt, line 3: 123

kirillt (commented):
will approved when all comments resolved

Issue #137, opened 2021-11-19
Context menu should be displayed by clicking at a tag


kirillt:
We should be able to run various actions involving tags, e.g. #49, #50.
The list of actions can be shown via context menu:
1. In Gallery screen we would just call it by single-tap.
    Current _remove tag from resource_ action (triggered by long-tap) must be moved into the menu.

2. We need to be able to call the menu from tag selector (in resource grid).
    Single-tap still should be handled as inclusion of clicked tag into the query.
    Double-tap must spawn context menu with options including _inverted selection_.

kirillt:
#196 should be also added into Gallery's context menu

PR #247, opened 2022-05-21
#137: Tag context menu

mdrlzy:
#137

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -67,4 +67,6 @@
     <string name="skipping_folder">Skipping %1$s</string>
     <string name="enough">Enough</string>
     <string name="access_denied">Access denied</string>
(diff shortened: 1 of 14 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -66,4 +66,7 @@
     <string name="scanning">Scanning</string>
     <string name="skipping_folder">Skipping %1$s</string>
     <string name="enough">Enough</string>
+    <string name="select">Select</string>
```
kirillt, line 69: Maybe better to call it "Navigate"? Because "select" sounds like it's possible to select more than one tag. Also "Remove" is probably better to rename into "Untag", because no data is really deleted.

kirillt (approved):
Looks awesome. Minor UX improvements will be needed, but we can do them later.
Also, actions in Gallery tag menu should be named differently.

Issue #116, opened 2021-10-26
Support plain text files


kirillt:
The app should support previewing plain text files (`.txt`) and editing them in an external app.

kirillt:
Open question: how should we determine if a resource is "plain text"? Ideally, we would like to treat not only files with `.txt` extension. Is it a good idea to take first _N_ bytes, attempt conversion of them to UTF-8 and, in case of success, treat the file as plain text?

mdrlzy:
I think we can use [Apache Tika](https://tika.apache.org/)
`val fileType = Tika().detect(inputStream)`

PR #263, opened 2022-06-11
#116: Added preview and thumbnail for txt file type.

pgpathan22:
#116


app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/generator/TxtPreviewGenerator.kt
```diff
@@ -0,0 +1,58 @@
...
+import android.text.Layout
+import android.text.StaticLayout
+import android.text.TextPaint
+import java.io.FileReader
+import java.nio.file.Path
+import space.taran.arknavigator.ui.App
+
+object TxtPreviewGenerator {
+    // it is padding in preview image
+    val padding = 10f * App.instance.resources.displayMetrics.density
```
kirillt, line 16: Do we need this for thumbnails too?

kirillt (commented):
Good work, minor changes needed.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
```diff
@@ -43,12 +45,16 @@ object PreviewGenerators {
...
         }
-
-        generatorsByExt[extension(path)]?.let { generator ->
+        val ext = extension(path)
+        generatorsByExt[ext]?.let { generator ->
             val time2 = measureTimeMillis {
-                val preview = generator(path)
-                val thumbnail = resizePreviewToThumbnail(preview)
-                storePreview(previewPath, preview)
+                val thumbnail = if (ext == "txt") {
```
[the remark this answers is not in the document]
pgpathan22: then we need to update the ` val ext = extension(path)` method to determine text file whether there is ext or not. If file without ext will determine using the Tika API and return ext as txt.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
```diff
@@ -43,12 +45,16 @@ object PreviewGenerators {
...
+                    resizePreviewToThumbnail(generator(path))
```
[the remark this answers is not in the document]
pgpathan22: ok. I will do this now.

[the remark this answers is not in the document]
pgpathan22: `Bitmap.createBitmap(..)` I have to add thumbnail size + padding value in width to set margin at right side of text.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/generator/TxtPreviewGenerator.kt
pgpathan22: Yes we need this for thumbnail. It is to add padding around the text in image.

mdrlzy (commented):
Looks nice. Only `PreviewItemViewHolder` looks confusing. I would refactor it after you finish.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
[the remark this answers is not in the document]
pgpathan22: Sure, will do it.

app/build.gradle
```diff
@@ -133,6 +133,10 @@ dependencies {
     androidTestImplementation "androidx.test:rules:1.4.0"
     androidTestImplementation "androidx.test.espresso:espresso-core:3.4.0"
     implementation 'com.github.wseemann:FFmpegMediaMetadataRetriever:1.0.14'
+
+    implementation 'org.apache.tika:tika-core:2.4.0'
+    //implementation 'org.apache.tika:tika-parsers-standard-package:2.4.0'
```
kirillt, line 138: What `tika-parsers-standard-package` could be useful for?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/ResourceKindFactory.kt
```diff
@@ -3,12 +3,15 @@ package space.taran.arknavigator.mvp.model.repo.kind
...
+import space.taran.arknavigator.utils.getMimeTypeUsingTika
 import java.nio.file.Path
 
 interface ResourceKindFactory<T : ResourceKind> {
     val acceptedExtensions: Set<String>
+    val acceptedMimeTypes: Set<String>
     val acceptedKindCode: KindCode
-    fun isValid(path: Path) = acceptedExtensions.contains(extension(path))
+    fun isValid(path: Path) = acceptedExtensions.contains(extension(path)) ||
+        acceptedMimeTypes.contains(getMimeTypeUsingTika(path))
```
kirillt, line 14: Maybe we should log warning if MIME type conflicts with extension.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/ResourceKindFactory.kt
```diff
@@ -37,9 +40,9 @@ object GeneralKindFactory {
     ): ResourceKind? {
         kindCode ?: return null
 
-        val data = extras.map {
+        val data = extras.associate {
```
kirillt, line 43: :+1:

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
```diff
@@ -89,3 +90,7 @@ fun reifySorting(sorting: Sorting): Comparator<ResourceMeta>? =
 
 const val KILOBYTE = 1024
 const val MEGABYTE = 1024 * KILOBYTE
+
+fun getMimeTypeUsingTika(path: Path): String? {
+    return Tika().detect(Files.newInputStream(path,), "image")
```
kirillt, line 95: This code can be important for performance. Is content of the file being read completely to the end or just first bytes?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
```diff
@@ -45,26 +47,64 @@ object PreviewGenerators {
...
-                val thumbnail = if (ext == "txt") {
-                    resizePreviewToThumbnail(generator(path))
-                } else {
-                    val preview = generator(path)
-                    storePreview(previewPath, preview)
-                    resizePreviewToThumbnail(preview)
-                }
+                val preview = generator(path)
+                storePreview(previewPath, preview)
+                val thumbnail = resizePreviewToThumbnail(preview)
```
kirillt, line 56: We shouldn't call this for text. Text should fall into the same branch as images (images don't have previews, the same as text resources). I think code above should be adapted to include text too:
```
if (ImageKindFactory.isValid(path)) { // || PlainTextKindfactory.isValid(path)
...
}
```

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
```diff
@@ -45,26 +47,64 @@ object PreviewGenerators {
...
+            return
         } ?: Log.d(
             PREVIEWS,
             "No generators found for type .${extension(path)} ($path)"
         )
+        // This section called when the code will failed to generate preview for
+        // pdf, link, txt extension.This code will match with existing factories'
+        // acceptedExtensions as well as acceptedMimeTypes (if ext is blank).
+        if (PlainTextKindFactory.isValid(path)) {
+            generatorsByExt["txt"]?.let { generator ->
```
kirillt, line 72: If you need to use this generator directly, just initialize it separately, save it both to map and variable and use the variable here. Don't use String constants for accessing it.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
```diff
@@ -45,26 +47,64 @@ object PreviewGenerators {
...
+                val preview = generator(path)
+                storePreview(previewPath, preview)
+                val thumbnail = resizePreviewToThumbnail(preview)
                 storeThumbnail(thumbnailPath, thumbnail)
             }
             Log.d(
                 PREVIEWS,
                 "Preview and thumbnail generated for $path in $time2 ms"
             )
+            return
```
kirillt, line 63: I think we should do better dispatch based on extension and Tika. Don't copypaste the code. The code below can be removed if you check necessary conditions before and call necessary sections of code just 1 time.

kirillt (commented):
Some more polishing needed.

app/build.gradle
pgpathan22: Not needed for now, I have deleted it.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/ResourceKindFactory.kt
pgpathan22: Currently we get conflict for svg file without extension.  So please tell me the case of the conflict for others ?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
pgpathan22: If we put this then I have to check again for txt file so I can called `            TxtPreviewGenerator.generate(path, THUMBNAIL_SIZE)
` to create bitmap using some text from the file. instead this, I will put `PlainTextKindfactory.isValid(path)` in else if condition.

app/src/main/java/space/taran/arknavigator/utils/FileUtils.kt
pgpathan22: Screenshot from 2022-06-17 14-19-48
Looks like the performance is managed by Tika.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
pgpathan22: Resolved.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
pgpathan22: Resolved.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/ResourceKindFactory.kt
kirillt: We can check it in generic way, something like this:
```
val extensionIsValid = acceptedExtensions.contains(extension(path))
val mimeTypeIsValid = acceptedMimeType.contains(getMimeTypeUsingTika(path))
if (extensionIsValid != mimeTypeIsValid) {
    log.warn("File extension is valid? ${extensionIsValid}")
    log.warn("MIME type is valid? ${mimeTypeIsValid}")
}
```

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewGenerators.kt
```diff
@@ -59,6 +75,33 @@ object PreviewGenerators {
             PREVIEWS,
             "No generators found for type .${extension(path)} ($path)"
         )
+        // This section called when the code will failed to generate preview for
+        // pdf, link, txt extension.This code will match with existing factories'
+        // acceptedExtensions as well as acceptedMimeTypes (if ext is blank).
+        if (
+            ext.isBlank() && getMimeTypeUsingTika(path = path) == "application/pdf"
+        ) {
+            generatorsByExt["pdf"]?.let { generator ->
```
kirillt, line 84: Looks like code duplication again, code duplication should be avoided at all cost, because it's difficult to maintain (easy to modify one place and forget about another). Maybe we need to create additional map `generatorByMIME`, fill it with the same values of `generatorByExt` just for different keys, check one map, then another, then invoke necessary code one time instead of several.

Also, no generator should be retrieved directly by mentioning String constant, like `"pdf"`. The maps should be used with variables, otherwise we can just store pdf generator into class constant and use it instead of searching in the map.

pgpathan22: Ok. I will change and call the method direct as it will more useful.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/ResourceKindFactory.kt
```diff
@@ -60,4 +62,15 @@ object GeneralKindFactory {
                 ResourceExtra(id, entry.key.ordinal, entry.value!!)
             }
     }
+
+    private fun findFactory(path: Path): ResourceKindFactory<ResourceKind>? {
```
kirillt, line 66: :+1:

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/ResourceKindFactory.kt
```diff
@@ -60,4 +62,15 @@ object GeneralKindFactory {
...
+        var factory = factories.find { it.isValid(path) }
+        if (factory != null) return (factory as ResourceKindFactory<ResourceKind>)
```
kirillt, line 68: Aha.. So we are calling Tika only if we haven't found any factory for the resource basing on its path. That's good, no need to worry about performance (I think reading first 100K characters is a bit demanding).

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/GeneralPreviewGenerator.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+        val time = measureTimeMillis {
+            generator.generate(path, previewPath, thumbnailPath)
+        }
+        Log.d(PREVIEWS, "Preview and thumbnail generated for $path in $time ms")
+    }
+
+    private fun findGenerator(path: Path): PreviewGenerator? {
+        var generator = generators.find { it.isValid(path) }
+        if (generator != null) return generator
```
kirillt, line 38: Aha.. So we are calling Tika only if we haven't found any generator for the resource basing on its path. That's good, no need to worry about performance (I think reading first 100K characters is a bit demanding).

kirillt (approved):
Good work, guys.

kirillt:
Please, sync with `main` branch to enable crash reporting in this PR.

Issue #50, opened 2021-09-01
Bulk resources operations


kirillt:
It should be possible to apply some operations to selections, e.g. to move all selected resources to some folder (_favorite_ or _root_) with all their tags.

kirillt:
Selection should be possible by either:
* specifying resources by one with long-tap
* using tag pop-up menu (#137)

kirillt:
Operations to support:
* sharing
* moving to another folder, migrate tags as well
* deletion
* labeling with a tag (maybe it's even possible to remove a tag shared among all of selected resources)

kirillt:
#281 implement core of this functionality
But we still need:
> sharing
> labeling with a tag

Advanced feature (not sure if it's needed):
> remove a tag shared among all of selected resources

Also we need to double-check later that tags migrate with resources into another root.

kirillt:
Continuation of work: https://github.com/ARK-Builders/ARK-Navigator/issues/292

PR #281, opened 2022-07-22
#50: Bulk resources operations

mdrlzy:
#50

### DIFF
--- a/app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
@@ -113,7 +113,7 @@ class ResourcesPresenter(
             tagsSelectorPresenter.calculateTagsAndSelection()
 
             val path = (rootAndFav.fav ?: rootAndFav.root)
-            val title = if (path != null) "$path, " else ""
+            val title = if (path != null) "${path.last()}, " else ""
 
             viewState.setToolbarTitle("$title${roots.size} of roots chosen")
             viewState.setProgressVisibility(false)
--- a/app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
@@ -1,7 +1,10 @@
 package space.taran.arknavigator.mvp.presenter
 
 import android.util.Log
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import moxy.MvpPresenter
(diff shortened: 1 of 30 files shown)


app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
```diff
@@ -113,16 +119,74 @@ class ResourcesPresenter(
...
-        super.onDestroy()
+    fun onMoveSelectedResourcesClicked(
+        directoryToMove: Path
+    ) = presenterScope.launch(Dispatchers.IO) {
+        withContext(Dispatchers.Main) {
+            viewState.setProgressVisibility(true, "Moving")
+        }
+        val resourcesToMove = gridPresenter.resources.filter { it.isSelected }
+        val results = resourcesToMove.map { item ->
+            async {
```
kirillt, line 137: Am I right, that would spawn a coroutine for each resource?

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
```diff
@@ -113,16 +119,74 @@ class ResourcesPresenter(
...
+                val path = index.getPath(item.meta.id)
+                val newPath = directoryToMove.resolve(path.name)
+                path.copyTo(newPath)
+                if (path != newPath)
+                    path.deleteIfExists()
+            }
+        }
+        results.forEach { it.await() }
+        index.reindex()
```
kirillt, line 146: Can we optimize this, telling into index what was moved and updating only affected paths? There should be no `ResourceId` re-calculation, in fact.

kirillt:
We also need _sharing_ and _labeling_ operations for selections.

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
mdrlzy: Yes, we spawn a coroutine for each resource. The only performance limit is the number of threads in the pool, since the operations block the thread.
Coroutines are really lightweight. We can easily spawn 100,000 coroutines.
https://kotlinlang.org/docs/coroutines-basics.html#coroutines-are-light-weight

app/src/main/java/space/taran/arknavigator/mvp/presenter/ResourcesPresenter.kt
mdrlzy: Id calculation does not occur. We only check files for existence and modification date.

mdrlzy: However, Id recalculation can happen when moving resources to another folder in the same root. It can be fixed

kirillt (approved):
Cool :rocket:

Issue #254, opened 2022-06-02
Metadata and preview (persisted/replicated) storages


kirillt:
Written for _Link-to-Web_ resources, but might be valid for other kinds as well.

Right now, we are storing metadata and preview in the same archive with the URL. Any change in metadata like title or description will change the link's _id_ and invalidate tags for new version. We can create dedicated metadata storage, like `.ark/meta`. This way it would be also easier to work with the same resource on several devices since we could modify tags on device A and modify metadata on device B, but both modification would point on the same _id_.

Previews also should be kept separately, because the same URL can get different preview just because of changes at remote site. If preview changes, we would just update picture in `.ark/previews` without need to update ids.

kirillt:
We need also "our own" filenames, which we would put into `.ark/meta`. Such _resource names_ would be de-coupled from _locators_ (filenames and paths). Since we use _content addressing_, the only locator we accept is `ResourceId` (and _root_ path, but this is temporary measure).

### E.g. we copy resources
```
(name=A, id=1),
(name=B, id=2),
(name=C, id=3)
```
from root **X** into root **Y**.

Root **Y** contains resources
```
(name=A, id=4),
(name=D, id=5),
(name=E, id=2)
```

#### Currently existing file browsers would solve this case like this:
```
File A exists in folder Y, do you want to overwrite? Y/N
```
What we have after the operation:
```
(name=A, id=1 or id=4),
(name=B, id=2),
(name=C, id=3),
(name=D, id=5),
(name=E, id=2)
```
Here we have:
1) less resources than we wanted: only 4 instead of 5 (we can't have both id=1 and id=4 under the same filename)
2) duplicates: two files with the same content but different _locators_

#### Resource browsers based on content-addressing should solve this case this way:
```
(No questions asked, user clearly specified which resource they want in destination)
```
What we have after the operation:
```
(name=A, id=1),
(name=[B,E], id=2),
(name=C, id=3)
(name=A, id=4),
(name=D, id=5)
```
I.e. we have all 5 unique resources, no questions asked. The only slight inconvenience we have is that one resource got two names. But we don't present this resource like different files, user can always choose name they like more or leave it as-is.

kirillt:
Previews storage is implemented now, we need also to create metadata storage.

PR #313, opened 2022-10-16
#254: Add metadata storage

hhio618:
### Description
#254 

### TODO
- [X] Add the Metadata storage
- [x] Add required `arklib` functions if needed (not required by this PR)
- [x] Read link files using the Metadata storage

### DIFF
--- a/app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourceMeta.kt
+++ b/app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourceMeta.kt
@@ -5,12 +5,16 @@ import space.taran.arknavigator.mvp.model.repo.kind.GeneralKindFactory
 import space.taran.arknavigator.mvp.model.repo.kind.MetaExtraTag
 import space.taran.arknavigator.mvp.model.repo.kind.ResourceKind
 import space.taran.arknavigator.mvp.model.repo.meta.MetadataStorage
+import space.taran.arknavigator.mvp.model.repo.meta.MetadataStorageRepo
+import space.taran.arknavigator.mvp.model.repo.preview.PreviewStorage
+import space.taran.arknavigator.mvp.model.repo.preview.PreviewStorageRepo
 import space.taran.arknavigator.utils.MetaResult
 import space.taran.arknavigator.utils.extension
 import java.io.IOException
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.attribute.FileTime
+import javax.inject.Inject
 
 
 data class ResourceMeta(
@@ -19,13 +23,11 @@ data class ResourceMeta(
     val extension: String,
     val modified: FileTime,
     val size: Long,
-    val kind: ResourceKind?,
-    val extras: Map<MetaExtraTag, String>?
+    var kind: ResourceKind?,
 ) {
-
     companion object {
 
-        fun fromPath(path: Path): MetaResult {
+        fun fromPath(path: Path, metadataStorage: MetadataStorage): MetaResult {
             val size = Files.size(path)
             if (size < 1) {
                 return MetaResult.failure(IOException("Invalid file size"))
@@ -33,27 +35,22 @@ data class ResourceMeta(
 
             val id = computeId(size, path)
 
-            var kind: ResourceKind? = null
-            var kindDetectException: Exception? = null
-
-            try {
-                kind = GeneralKindFactory.fromPath(path)
-            } catch (e: Exception) {
-                kindDetectException = e
-            }
-
             val meta = ResourceMeta(
                 id = id,
                 name = path.fileName.toString(),
                 extension = extension(path),
                 modified = Files.getLastModifiedTime(path),
                 size = size,
-                kind = kind,
-                extras = null,
+                kind = null
             )
 
-            val extras =
+            var kindDetectException: Exception? = null
 
+            try {
+                meta.kind = GeneralKindFactory.fromPath(path, meta, metadataStorage)
+            } catch (e: Exception) {
+                kindDetectException = e
+            }
             return MetaResult(meta, kindDetectException)
         }
 
@@ -64,8 +61,7 @@ data class ResourceMeta(
                 extension = room.resource.extension,
                 modified = FileTime.fromMillis(room.resource.modified),
                 size = room.resource.size,
-                kind = GeneralKindFactory.fromRoom(room.resource.kind, room.extras),
-                extras = room.extras.associate { MetaExtraTag.values()[it.ordinal] to it.value }
+                kind = GeneralKindFactory.fromRoom(room.resource.kind, room.extras)
             )
     }
 }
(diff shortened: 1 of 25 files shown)


app/src/main/java/space/taran/arknavigator/mvp/model/repo/meta/PlainMetadataStorage.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    override fun locate(path: Path, resource: ResourceMeta): Map<MetaExtraTag, String?>? {
+        val metadata = metaPath(resource.id)
+        if (!Files.exists(metadata)) {
+            Log.w(METADATA, "metadata was not found for resource $resource")
+            // means that we couldn't generate anything for this kind of resource
+            return null
+        }
+        val json = Json.parseToJsonElement(metadata.bufferedReader().use { it.readText() }).jsonObject
```
kirillt, line 44: Am I right that we encode `extras` field of `ResourceMeta` as JSON and write it in `.ark/meta/<resource id>`?

This should be good approach, just want to make sure I understand it correctly.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourceMeta.kt
```diff
@@ -2,21 +2,25 @@ package space.taran.arknavigator.mvp.model.repo.index
...
+
 data class ResourceMeta(
     val id: ResourceId,
     val name: String,
     val extension: String,
     val modified: FileTime,
     val size: Long,
-    val kind: ResourceKind?
+    val kind: ResourceKind?,
+    val extras: Map<MetaExtraTag, String>?
```
kirillt, line 23: We've just realized with Dinar that this slightly wrong, because we already have sub-structures in `ResourceKind` class. Yes, the name is slightly wrong, because `ResourceKind` is not plain enumeration type, but contains sub-structures inside each branch. Is it possible to re-design this PR in order to remove `extras` and use `ResourceKind` internal data instead?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/meta/PlainMetadataStorage.kt
hhio618: Yeah, that's right!

app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourceMeta.kt
hhio618: NP, I'll make the required changes.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/meta/PlainMetadataStorage.kt
```diff
@@ -34,24 +23,25 @@ class PlainMetadataStorage(val root: Path) : MetadataStorage {
         metaDir.createDirectories()
     }
 
-    override fun locate(path: Path, resource: ResourceMeta): Map<MetaExtraTag, String?>? {
+    override fun locate(path: Path, resource: ResourceMeta): ResourceMeta? {
         val metadata = metaPath(resource.id)
         if (!Files.exists(metadata)) {
             Log.w(METADATA, "metadata was not found for resource $resource")
             // means that we couldn't generate anything for this kind of resource
```
kirillt, line 30: Probably we should treat this case as that we don't have extra metadata (that is "metadata" for a user), but still return `resource: ResourceMeta` parameter as a result, since we can't have a resource without metadata (from systemic view). Lack of `kind` would mean the resource is of unknown type. That would convert return type from `ResourceMeta?` to `ResourceMeta`.

@mdrlzy What do you think?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/LinkKindFactory.kt
```diff
@@ -4,22 +4,23 @@ import kotlinx.serialization.Serializable
...
-    override fun fromPath(path: Path): ResourceKind.Link {
-        val linkJson = loadLinkFile(path.pathString)
-        val link = Json.decodeFromString(JsonLink.serializer(), linkJson)
+    override fun fromPath(path: Path, meta: ResourceMeta, metadataStorage: MetadataStorage): ResourceKind.Link {
+        // TODO: we don't need these lines anymore, but what about the native bindings?
+        // val linkJson = loadLinkFile(path.pathString)
+        // val link = Json.decodeFromString(JsonLink.serializer(), linkJson)
 
-        return ResourceKind.Link(link.title, link.desc, link.url)
+        return metadataStorage.locate(path, meta) as ResourceKind.Link
```
kirillt, line 23: Probably we don't need `fromPath` method at all.. Is it a good idea to just use `metadataStorage.locate` instead? @mdrlzy

app/src/main/java/space/taran/arknavigator/mvp/model/repo/index/ResourceMeta.kt
kirillt: Awesome, thanks :+1:

app/src/main/java/space/taran/arknavigator/mvp/model/repo/meta/PlainMetadataStorage.kt
```diff
@@ -0,0 +1,56 @@
...
+            ResourceKind.serializer(),
+            metadata.bufferedReader().use { it.readText() })
+        return result
+    }
+
+    override fun forget(id: ResourceId) {
+        metaPath(id).deleteIfExists()
+    }
+
+    override fun generate(path: Path, meta: ResourceMeta) {
```
kirillt, line 44: Probably `store` would be a better name for this name, since the data itself is already generated and we only need to write it to disk.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/kind/LinkKindFactory.kt
mdrlzy: We need it for now, because for the rest of `kind` we don't yet generate a json file.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/meta/PlainMetadataStorage.kt
hhio618: Done!

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewStorage.kt
```diff
@@ -10,5 +10,5 @@ interface PreviewStorage {
 
     fun forget(id: ResourceId)
 
-    fun generate(path: Path, meta: ResourceMeta)
+    fun store(path: Path, meta: ResourceMeta)
 }
```
mdrlzy, line 14: I don't think it needs to be renamed.

mdrlzy (commented):
Overall looks ok

app/src/main/java/space/taran/arknavigator/mvp/model/repo/preview/PreviewStorage.kt
hhio618: NP, what about this comment: https://github.com/ARK-Builders/ARK-Navigator/pull/313#discussion_r998221579
@kirillt

kirillt: @mdrlzy yeah, not really an issue of course, but IMO `generate` sounds like we generate a data, but here we only store existing data by path (we generate a file, but not the data).

kirillt (approved):

Issue #314, opened 2022-10-22
Bounty #1 for Shubert


kirillt:
Here are 2 issues we want to solve:

1) https://github.com/ARK-Builders/ARK-Navigator/issues/15
Shuffling current selection must result in random order of resources every time the button is pressed.

2) https://github.com/ARK-Builders/ARK-Navigator/issues/14
Scoring buttons and scores storage.

HLABZcc:
Interested 😄

gitcoinbot:
Issue Status: 1. Open 2. Started 3. **Submitted** 4. Done 

<hr>

__Work for 6826.3879 ZIL (200.00 USD @ $0.03/ZIL) has been submitted by__:

 1. [@shubertmunthali](https://gitcoin.co/shubertmunthali)



@kirillt please take a look at the submitted work:
* [PR](https://github.com/ARK-Builders/ARK-Navigator/pull/316) by @shubertmunthali

<hr>

* Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/29475)
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/29475).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
* $1,003,893,805.03 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

kirillt:
Bounty is paid: https://viewblock.io/zilliqa/tx/0x7635f2258ab4bcb0d3290d973f819f1a5d44228e02e9d60778e6468f18bf3e09
Gitcoin's integration with Zilliqa is outdated and not gonna be fixed for a while.

Issue #15, opened 2021-08-12
"Shuffle" button in resources grid


kirillt:
The button must shuffle current selection every time (it isn't "as-is" ordering).

Lazy shuffling could be implemented for usage with big collections: select random N resources, where N is amount of visible items in the grid; select next ones only when a user scrolls the grid or the gallery. Similar to #73.

kirillt:
Implemented in #316

Issue #14, opened 2021-08-12
Score of a resource and "boost" buttons


kirillt:
Sometimes it is convenient to pin a resource, so it would be shown on top of other resources. Score is more general approach, which allows to prioritize one resource over another and also to deprioritize resources, so resources with negative score would be shown in the bottom of current selection.

Scoring requires separate storage similar to tags storage, but much simpler — file `.ark/scores` should contain mapping from Resource ID into integer number (can be positive or negative). Default value should be 0, so if a resource is not mentioned in the file, that implicitly means it's score is 0.

For modifying score of a resource, "boost" buttons should be drawn on top of its preview in GalleryFragment. Just "+" and "-".

The app should have switch in resources grid, enabling/disabling usage of score during resources sorting.

kirillt:
Implemented in #316

PR #316, opened 2022-11-02
#314: Resource shuffling and pinning

ShubertMunthali:
Issues #14 #15
Bounty #314


kirillt:
The "shuffle" button looks nice, I waited for this feature more than a year actually. Will be useful for learning languages.

UX comments:
- [ ] I described this poorly before, but we need quick `+` and `-` buttons in Gallery mode for increasing/decreasing score, arrows up/down also could be used as icons for these buttons. We still need action in "selection menu" (in Resources Grid screen), but it should be not "pin", but increase/decrease again and also "reset score" which erases selected resources' ids from scores storage.
- [ ] I realized that only now, that we need 2 states of "shuffle" button: "pressed" and "un-pressed". Default state is "un-pressed". When a user taps on it, it becomes "pressed" and resources are shuffled. When the user presses it second time, the button must become "un-pressed" and resources must become ordered again (according to chosen sorting). Otherwise, user would need work around this by changing tags in order to see ordered selection again.

kirillt:
@mdrlzy Could you guide Shubert regarding how to extract common logic of the storages into re-usable module/class? Also Shubert might need help/guidance with unit tests.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/ScoreStorage.kt
```diff
@@ -0,0 +1,17 @@
...
+interface ScoreStorage {
+
+    fun contains(id: ResourceId): Boolean
+
+    fun setScore(id: ResourceId, score: Score)
+
+    fun getScore(id: ResourceId): Score?
+
+    fun countScores(): Int
+
```
mdrlzy, line 15: Why do we need this function?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/ScoreStorage.kt
```diff
@@ -0,0 +1,17 @@
...
+import space.taran.arknavigator.utils.Score
+
+interface ScoreStorage {
+
+    fun contains(id: ResourceId): Boolean
+
+    fun setScore(id: ResourceId, score: Score)
+
+    fun getScore(id: ResourceId): Score?
+
```
mdrlzy, line 13: I think you can return non-nullable type since it can only be null in aggregated mode, but we have `contains` function for check

mdrlzy:
**Bug:** Pinned files stop being first if you change sorting

<details>
  <summary>Screenshot</summary>

photo_2022-11-04_17-58-24
</details>

mdrlzy:
Actually scores and pinned are two different things.
Looks like we need a separate storage for pinned ones.
@kirillt What do you think?

kirillt:
@mdrlzy yeah, pinning is different although looks nice. We could implement pinning using scores if we allow setting `MAX` score :) Both with pinning and scores, ordering must affect only positions of resource with the same scores.

In other words, first we sort by score (always when scores are enabled, we need a button or checkpoint for this), then we sort by the selected criteria.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/ScoreStorage.kt
ShubertMunthali: It was being used to find a pinning position as a score but now that we have changed. It's no longer needed. It will be omitted

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/ScoreStorage.kt
ShubertMunthali: Yeah, and I realise it always returns a value

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/PlainScoreStorage.kt
```diff
@@ -53,23 +53,14 @@ class PlainScoreStorage(
         if (!scoreById.containsKey(id))
             error("Storage isn't aware of this resource id")
 
-        if (score <= 0) {
```
kirillt: We allow negative scores indeed, but we should not store 0 scores since that could make the storage excessively big. `0` should be default value.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/PlainScoreStorage.kt
```diff
@@ -53,23 +53,14 @@ class PlainScoreStorage(
...
-
         Log.d(
             SCORES_STORAGE,
             "new score for resource $id: $score"
         )
         scoreById[id] = score
     }
 
-    override fun getScore(id: ResourceId) = scoreById[id]
+    override fun getScore(id: ResourceId) = scoreById[id]!!
```
kirillt, line 63: Here we should fallback to `0` as the default value.

kirillt:
Star icon looks cool!

I think this feature would be really nice for UX:
- [ ] Button to hide scores, all its controls and do normal sorting.

I've found a tricky bug:
- [ ] If you open a folder. Edit some picture (overwrite it). Return back to Navigator, stay in Gallery mode, click `+` then `-`. First, the score goes to `1`, then to `-1`, after you can't make it any other number, only `1` or `-1`

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
```diff
@@ -244,14 +310,24 @@ class ResourcesGridPresenter(
...
+        val selection = this.selection.toSet()
+        this.selection = resources.filter {
+            selection.contains(it)
+        }.sortedWith(
+            compareBy {
+                !it.isPinned
+            }
+        )
+    }
+
```
mdrlzy, line 330: I think it's better to rename it to `sortSelection`

mdrlzy (commented):
Looks nice

mdrlzy:
@kirillt I think it's a bad idea to use scores to determine pinned
For example, we define that `isPinned = score == Int.MAX_VALUE`
How to handle score increase/decrease in gallery? 
Do we need to prohibit changing score for pinned ones?

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/PlainScoreStorage.kt
ShubertMunthali: This is handled, no zero score is persisted

ShubertMunthali:
> @kirillt I think it's a bad idea to use scores to determine pinned For example, we define that `isPinned = score == Int.MAX_VALUE` How to handle score increase/decrease in gallery? Do we need to prohibit changing score for pinned ones?

I think it will be simpler to implement them separately if they can co-exist. I removed the pinning feature and added scores. We can add pinning later now.

app/src/main/java/space/taran/arknavigator/utils/ComparatorUtils.kt
```diff
@@ -16,3 +19,9 @@ fun <T> unequalCompareBy(comparator: Comparator<T>): Comparator<T> =
             )
         } else result
     }
+
+fun compareByScores(scoreStorage: ScoreStorage) = Comparator<ResourceItem> { a, b ->
```
kirillt, line 23: Better to move it somewhere closer to `ScoreStorage`, maybe even in the same module. Or to `ResourceGridPresenter` since it is used only there.

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
```diff
@@ -316,13 +313,21 @@ class ResourcesGridPresenter(
             if (comparator != null) {
                 resources = resources.map { it to it }
                     .toMap()
-                    .toSortedMap(unequalCompareBy(comparator))
+                    .toSortedMap(
+                        unequalCompareBy(
+                            if (sortByScores)
+                                compareResourcesByScores()
+                                    .then(comparator)
```
kirillt, line 320: :+1:

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
```diff
@@ -219,6 +236,21 @@ class GalleryPresenter(
...
+
+    private fun changeScore(inc: Score) = presenterScope.launch {
+        val score = scoreStorage.getScore(currentResource.id) + inc
+        scoreStorage.setScore(currentResource.id, score)
+        withContext(Dispatchers.IO) {
+            scoreStorage.persist()
+        }
+        withContext(Dispatchers.Main) {
+            viewState.displayScore(score)
+        }
```
mdrlzy, line 251: Here you also need to update scores on resources screen, for this you can use `viewState.notifyResourcesChanged()`

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
```diff
@@ -182,12 +206,72 @@ class ResourcesGridPresenter(
...
+        scoreStorage.resetScores(selectedResources)
+        withContext(Dispatchers.Main) {
+            sortAllResources()
+            sortSelectionAndUpdateAdapter()
+            onSelectingChanged(false)
+        }
+    }
+
+    fun allowScoring() =
+        selectedResources.isNotEmpty() && selectedResources.size == 1 && sortByScores
```
mdrlzy, line 247: Why do we disallow scoring for multiple resources?

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragmentHelper.kt
```diff
@@ -93,6 +94,26 @@ fun ResourcesFragment.setupAndShowSelectedResourcesMenu(menuBtn: View) {
...
+                popup.popupWindow.dismiss()
+            }
+        }
+        with(btnDecreaseScore) {
+            isVisible = presenter.allowScoring()
+            setOnClickListener {
+                presenter.onDecreaseScoreClicked()
+                popup.popupWindow.dismiss()
+            }
+        }
```
mdrlzy, line 110: I think popup can be left after changing score

app/src/main/java/space/taran/arknavigator/mvp/presenter/GalleryPresenter.kt
ShubertMunthali: I have implemented it

app/src/main/java/space/taran/arknavigator/mvp/presenter/adapter/ResourcesGridPresenter.kt
ShubertMunthali: I thought scoring should work for one resource every time. But that way, we can score many resources at once. I think that's ok

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragmentHelper.kt
ShubertMunthali: Oh! yeah, sounds nice

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
```diff
@@ -87,6 +87,8 @@ class ResourcesFragment : MvpAppCompatFragment(), ResourcesView {
 
     private var tagsSelectorAdapter: TagsSelectorAdapter? = null
 
+    private var shuffledState = -1
```
kirillt, line 90: Looks like it should be plain `Boolean` :)

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragmentHelper.kt
```diff
@@ -93,6 +94,45 @@ fun ResourcesFragment.setupAndShowSelectedResourcesMenu(menuBtn: View) {
...
+                    toast(R.string.select_at_least_one_resource)
+                    popup.popupWindow.dismiss()
+                    return@setOnClickListener
+                }
+                ConfirmationDialogFragment.newInstance(
+                    getString(R.string.are_you_sure),
+                    getString(
+                        R.string.resource_scores_erased,
+                        selectedSize,
+                        if (selectedSize == 1) "" else "s"
```
kirillt, line 128: :+1:

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragment.kt
ShubertMunthali: Boolean brings conflicts when being updated with clicks on the button

ShubertMunthali: I will find a better way to use it though :)

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/PlainScoreStorage.kt
```diff
@@ -0,0 +1,173 @@
...
+            return@withContext result
+        }
+
+    private suspend fun writeStorage() = withContext(Dispatchers.IO) {
+        val lines = mutableListOf<String>()
+
+        lines.add("$STORAGE_VERSION_PREFIX$STORAGE_VERSION")
+
+        val entries = scoreById.filterValues {
+            it > 0 || it < 0
```
kirillt, line 135: `it != 0`

ShubertMunthali: Oh yeah:)

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragmentHelper.kt
ShubertMunthali: Currently, we only have one language. If support for many languages is added we can always created the resources. I think for now that's simple.

app/src/main/java/space/taran/arknavigator/mvp/model/repo/scores/PlainScoreStorage.kt
ShubertMunthali: I have it resolved

app/src/main/res/layout/fragment_gallery.xml
```diff
@@ -98,6 +100,42 @@
...
+            app:layout_constraintTop_toBottomOf="@+id/remove_resource_fab"
+            app:srcCompat="@drawable/ic_arrow_up" />
+        <ImageView
+            android:id="@+id/starIcon"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            app:layout_constraintEnd_toEndOf="@+id/remove_resource_fab"
+            app:layout_constraintStart_toStartOf="@+id/remove_resource_fab"
+            app:layout_constraintTop_toBottomOf="@+id/increaseScore"
+            app:srcCompat="@drawable/ic_star" />
```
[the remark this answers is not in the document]
ShubertMunthali: Oh! Yeah! I forgot that one :). I will resolve it

app/src/main/java/space/taran/arknavigator/ui/fragments/ResourcesFragmentHelper.kt
kirillt: Let's not mess with it for this moment.

kirillt (approved):

Issue #283, opened 2022-07-22
Condensed folder picker


kirillt:
In some of bulk operations, like copying, we need to select a destination folder. For this, we present folder picker to the user, which contains all top-level folders on the device. We should provide user with clutter-free experience, which means only folders with his own data are displayed. We assume that if user uses ARK Navigator, then they added all of their data into it :)

So, we need separate kind of folder picker which displays only _root_ and _favorite_ folders. Essentially such a picker would repeat structure of Folders screen. It would be pragmatic to have possibility to expand condensed folder picker into normal one allowing the user to move the resources wherever they want.

Development of this idea would be separating Folders screen into dedicated app/service, keeping all known `.ark` folders and providing [folder picker](https://github.com/ARK-Builders/ark-filepicker) with data about user's content folders.

PR #318, opened 2022-12-03
#283: Condensed folder picker

mdrlzy:
#283


kirillt:
Update erased folders, I guess that is expected since we changed path of roots list storage, but:
- [ ] Auto-discovery was not suggested after the update.
- [ ] Manual addition of a root causes RoomDB exception:
http://65.21.4.137:8080/app/1/bug/95/report/93eb90c9-f755-493f-a684-12165d2cc6e2

After manual data erasure in settings menu:
- [ ] Auto-discovery triggers indexing of each root, which makes user wait too long.

mdrlzy:
> Auto-discovery was not suggested after the update.

Added rescan roots button to the settings

> Manual addition of a root causes RoomDB exception

I forgot to change database version, also added small manual migration to avoid destructive migration and unnecessary indexing

> After manual data erasure in settings menu:
> Auto-discovery triggers indexing of each root, which makes user wait too long.

You also deleted the app database, so you need to reindex roots

kirillt:
> You also deleted the app database, so you need to reindex roots

Yes, but we can do it lazily, waiting till particular root will be opened, isn't it? This way, user would only wait every time she opens a root for first time, but would not wait at all during opening the app.

> Added rescan roots button to the settings

:+1: 

> I forgot to change database version, also added small manual migration to avoid destructive migration and unnecessary indexing

Will test update again.

kirillt (approved):

PR #325, opened 2023-01-12
Migrating resource index to ArkLib-Android

ShubertMunthali:


kirillt:
@ShubertMunthali could you test this PR with this test data?
[test.zip](https://github.com/ARK-Builders/ARK-Navigator/files/10439547/test.zip)

It must open and have several tags and scores set.

kirillt:
Manual migration of tags and scores to new version or `ResourceId` (Linux):
```
1. cd /path/to/root   # this is a root to migrate
2. ark-cli monitor    # creates .ark/index
3. (for new in $(cat .ark/index | cut -f2 -d' '); do old=$(echo $new | sed 's/..-//g'); echo "s/$old/$new/"; done) > .ark/migration
4. while read -r pattern ; do sed $pattern -i .ark/{tags,scores} ; done < .ark/migration
```

kirillt:
Now the app:
1. Doesn't open migrated tags and scores, although syntax looks valid.
2. Doesn't preserve new tags or scores. If you set some and restart the app, you'll see empty tags and no scores.

Here is a test root to check against:
[test.zip](https://github.com/ARK-Builders/ARK-Navigator/files/10439715/test.zip)
It already contains ids in new format (`size-crc32`).

@ShubertMunthali could you check?

ShubertMunthali:
I will look into that

kirillt (approved):

Issue #153, opened 2021-12-07
Faster PDF preview generation


kirillt:
Right now, we use https://github.com/barteksc/PdfiumAndroid for rendering PDF previews.
The library is loaded per each document instead of loading it once per batch,
because it doesn't have API to re-use instance several times.

We should consider alternatives:
https://pspdfkit.com/guides/android/viewer/kotlin/
https://developer.android.com/reference/android/graphics/pdf/PdfRenderer.html
https://github.com/pdf-rs/pdf

kirillt:
Testing data:
https://www.taran.space/dev/TEST_PDF.zip

kirillt:
https://github.com/ARK-Builders/ARK-Navigator/pull/329

PR #271, opened 2022-07-10
#153: Faster PDF preview generation

kirillt:
#153 
Adopting new PDF functionality from [arklib](https://github.com/ARK-Builders/arklib/)
See https://github.com/ARK-Builders/arklib/issues/11 and https://github.com/ARK-Builders/arklib-android/issues/7 for details

PR #329, opened 2023-01-22
#153: PDF preview generation with Rust

mdrlzy:


kirillt:
**One more benchmark**
OS: Android 11
Device: Samsung Galaxy Note 10+
Benchmark data: 477 PDF files, 857 MB total

Comparing:
* `main` branch
    build: https://github.com/ARK-Builders/ARK-Navigator/actions/runs/3931502951
* this PR version
   build: https://github.com/ARK-Builders/ARK-Navigator/actions/runs/3934199593

Parallel previews/thumbnails rendering is ON, that's why results are quicker than in https://github.com/ARK-Builders/ARK-Navigator/pull/271.

In this PR, PDF rendering is limited to be performed only for 1 PDF at a time, but thumbnails should still be scaled down from ready previews in parallel.

Results (total indexing time, previews generation is the bottleneck):
**main: 26 sec**
**this PR: 47 sec**

Accepting this PR because we value portability higher in this project and also performance penalty might be temporary.

We need to make https://github.com/ajrcarey/pdfium-render thread-safe :)

kirillt (approved):

Issue #334, opened 2023-02-16
Resource Grid UI enhancements


kirillt:
- [x] Dedicated switch for order of sorting
- [x] Dedicated scores switch in tag selector

![ui](https://user-images.githubusercontent.com/581023/219444651-eed11d3d-3d28-49ac-af1c-e188ed2e2d8c.png)

kirillt:
Implemented in #342 and #345

Issue #333, opened 2023-02-16
Initialize the app with popular folders


kirillt:
When a user opens the app for the very first time, empty folders list could be confusing.

We need to provide a user with option to initialize the app with default folders.
These folders could be:
* `DCIM/Camera`
* `Pictures`
* `Downloads`
* `Documents`

For this issue be considered solved, we need:
- [ ] determine clear list of "popular" folders
- [ ] determine variations of each path in different Android versions
- [ ] modify "first run dialog" in the way it would not only suggest to scan filesystem for previously created folders, but also to add "popular" folders

kirillt:
Additional requirements:
- [ ] folder suggesting dialog must be separate from scan dialog
- [ ] folder suggesting dialog must be presented after scan dialog
- [ ] if any of suggested folders is already added, it should be excluded from suggestions
- [ ] if no folders remain to suggest, folder suggesting dialog can be omitted

kirillt:
Also, improvements to scan dialog can be made:
- [ ] if the app can't find `.ark-global` folder with a list of roots inside, it should ask for scan
- [ ] if there is `.ark-global` folder with a list of roots inside, the app shouldn't ask for scan

Issue #332, opened 2023-02-16
Aggregated mode must be disabled when there are no roots


kirillt:
When there are no folders added on folders screen:
- [x] `[navigate]` button must be disabled, this must be clear by gray shade of main color
- [x] folders list must display hint "Add at least one folder to start using the app."

It would be good to also add assertion that folders list isn't empty into aggregated mode initialization.

In current state, user can enter aggregated mode with no folders added. This is a confusing state, the app can't be useful in this state.

kirillt:
Implemented in #342

Issue #144, opened 2021-11-20
Progress spinner improvements


kirillt:
1. During loading a folder, more granular labels than just "Indexing" are needed. Previews generation isn't really a part of indexing, even though it is triggered by it. So, we need also "Generating previews".

2. Total number and number of completed resources should be displayed.
![progress-counter_](https://user-images.githubusercontent.com/581023/182830407-1ba217ca-d559-468d-8e06-63f89be9b6a8.jpg)

kirillt:
Implemented in #343

PR #342, opened 2023-03-30
UI improve

saxihuangxing:
- [x] #332
- [x] #334 
- [ ] #144
- [ ] #333


kirillt:
#332 and #334 are fixed in this PR, here are 2 small UI bugs though:
- [ ] "Add at least one folder ..." label is seen shortly during app load, even if user already has several folders
- [ ] "Scores" label near the switch misses first letter

bug1
bug2

kirillt (approved):

PR #388, opened 2023-10-08
Refactor FileUtils & cover unit test

hieuwu:
### Analysis
Currently, the `FileUtils` hold many responsibilities that lead to inconvinience when cover Unit test. This file contains lots of extension functions related to `Path` class and call `App.instance` to get list of device paths.
This is not really a good design when refer to a singleton object inside method body, we can not mock the dependency when write test

### Solution
Separate `FileUtils` unto 2 parts. 1 is used to extract the list devices and 1 for extension of `Path` class

- `DevicePathsExtractor` and its implementation `DevicePathsExtractorImpl`. 
- PathExt.kt

### Result
DevicePathsExtractorImpl
Screenshot 2023-10-08 at 22 51 17

PathExt
Screenshot 2023-10-08 at 22 50 41


.gitignore
```diff
@@ -12,3 +12,4 @@
 *.tab
 *.tab.values.at
 misc.xml
+/htmlReport
```
hieuwu, line 15: This is to ignore generated files when run test with coverage locally

app/src/main/java/dev/arkbuilders/navigator/data/utils/PathExt.kt
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    if (paths.size == 1) {
+        return paths.first()
+    }
+
+    return tailrec(ROOT_PATH, paths).first
+}
+
+private fun tailrec(_prefix: Path, paths: List<Path>): Pair<Path, List<Path>> {
```
kirillt, line 47: We need another name for `_prefix`... Doesn't `_` usually mean that the variable isn't actually used?

app/src/test/java/dev/arkbuilders/navigator/data/utils/PathExtKtTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import kotlin.io.path.notExists
+
+@RunWith(MockitoJUnitRunner::class)
+class PathExtKtTest {
+
+    private lateinit var testee: Path
+
+    @BeforeEach
+    fun setUp() {
+        testee = Paths.get("notAndroid")
```
kirillt, line 23: But what does `notAndroid` mean?

app/src/test/java/dev/arkbuilders/navigator/data/utils/PathExtKtTest.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        every { destPath.fileName } returns Paths.get("notAndroid")
+        every { sourcePath.resolve(Paths.get("notAndroid")) } returns Paths.get("notAndroid")
+        every { sourcePath.resolve(Paths.get("notAndroid")).notExists() } returns false
+
+        every { sourcePath.resolve("notAndroid_1.") } returns Paths.get("notAndroid_1.")
+
+        val result = sourcePath.findNotExistCopyName(destPath)
+
+        assertEquals("notAndroid_1.", result.name)
```
kirillt, line 39: Maybe just `"A"` and `"B"` would be easier to read instead of `notAndroid` and `not_Android_1.`..

app/src/main/java/dev/arkbuilders/navigator/data/utils/PathExt.kt
hieuwu: Argument name should start with lower case character. I updated it and also rename others

app/src/test/java/dev/arkbuilders/navigator/data/utils/PathExtKtTest.kt
hieuwu: I would like to use `PATH` instead. It's more meaningful

app/src/test/java/dev/arkbuilders/navigator/data/utils/PathExtKtTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    @Test
+    fun givenPath_whenFindNotExistCopyNameFalse_thenReturnNewPath() {
+        val sourcePath = mockk<Path>()
+        val destPath = mockk<Path>()
+
+        every { destPath.fileName } returns Paths.get("PATH")
+        every { sourcePath.resolve(Paths.get("PATH")) } returns Paths.get("PATH")
+        every { sourcePath.resolve(Paths.get("PATH")).notExists() } returns false
+
+        every { sourcePath.resolve("PATH_1.") } returns Paths.get("PATH_1.")
```
mdrlzy, line 35: You can test without a mock, use `createTempFile`

app/src/test/java/dev/arkbuilders/navigator/data/utils/DevicePathsExtractorTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        val directoryIterator: MutableIterator<Path> = arrayListOf(
+            Paths.get("PATH")
+
+        ).iterator()
+        every { mockedPath.iterator() } returns directoryIterator
+
+        testee.listDevices()
+
+        verify { mockedApplication.getExternalFilesDirs(null).toList() }
```
mdrlzy, line 48: Are we only testing here that `DevicePathsExtractorImpl` calls `App.getExternalFilesDirs(null).toList()`?
@kirillt Looks like overengineering

app/src/main/java/dev/arkbuilders/navigator/data/utils/DevicePathsExtractor.kt
```diff
@@ -0,0 +1,10 @@
+package dev.arkbuilders.navigator.data.utils
+
+import java.nio.file.Path
+
+enum class Sorting {
+    DEFAULT, NAME, SIZE, LAST_MODIFIED, TYPE
+}
+interface DevicePathsExtractor {
+    fun listDevices(): List<Path>
+}
```
mdrlzy, line 10: 👍

mdrlzy (commented):
After merge we need to move `Sorting enum`, `DevicePathsExtractor`, `DevicePathsExtractorImpl` to arklib

app/src/test/java/dev/arkbuilders/navigator/data/utils/PathExtKtTest.kt
hieuwu: Could you please elaborate this usage? Currently the tested method is interacting with `Path` objects. And createTempFile from File class returns File object instead.

mdrlzy: `createTempFile` returns `Path`

app/src/test/java/dev/arkbuilders/navigator/data/utils/DevicePathsExtractorTest.kt
mdrlzy: I mean, perhaps this test is not needed at all.
But let @kirillt decide

app/src/test/java/dev/arkbuilders/navigator/data/utils/PathExtKtTest.kt
hieuwu: It looks like it does not work. I would like to keep it as is. I also update result from the method execution

mdrlzy: What exactly doesn't work?

hieuwu: > What exactly doesn't work?

It works now. Thanks. Nice suggestion

app/src/test/java/dev/arkbuilders/navigator/data/utils/DevicePathsExtractorTest.kt
kirillt: Let's keep it for this moment.

app/src/test/java/dev/arkbuilders/navigator/data/utils/PathExtKtTest.kt
```diff
@@ -0,0 +1,88 @@
...
+                Paths.get("PATH"),
+                Paths.get("PATH2")
+            )
+        )
+        assertEquals("/", result.toString())
+    }
+
+    @Test
+    fun whenFindLongestCommonPrefixAndOnePrefixGroup_thenReturnFirstRelativePairOfPrefixAndPath() {
+        // TODO: Temporary test for unused method. Need to be improved for the method
```
kirillt, line 73: TODO

kirillt (approved):

PR #394, opened 2023-10-14
[build errors] Correct package name for classes from arklib due to re…

tuancoltech:
Correct package name for classes from arklib due to recent package name refactoring.
Please **DO NOT MERGE** until all necessary library versions are released as mentioned.


app/build.gradle
```diff
@@ -91,8 +91,11 @@ dependencies {
     def arklibVersion = '0.3.2'
 
     releaseImplementation "dev.arkbuilders:arklib:$arklibVersion"
-    debugImplementation "dev.arkbuilders:arklib-debug:$arklibVersion"
-    implementation("dev.arkbuilders:components:0.0.3") {
+    debugImplementation ("dev.arkbuilders:arklib-debug:$arklibVersion") {
+        //REMOVE below line after upgrading filepicker to latest version in arklib/arklib-debug
+        exclude group: "com.github.ARK-Builders", module: "ark-filepicker"
+    }
+    implementation("dev.arkbuilders:components:0.0.4-snapshot-01") {
```
tuancoltech, line 98: Change to new `components` version after releasing.

app/build.gradle
```diff
@@ -91,8 +91,11 @@ dependencies {
     def arklibVersion = '0.3.2'
 
     releaseImplementation "dev.arkbuilders:arklib:$arklibVersion"
-    debugImplementation "dev.arkbuilders:arklib-debug:$arklibVersion"
-    implementation("dev.arkbuilders:components:0.0.3") {
+    debugImplementation ("dev.arkbuilders:arklib-debug:$arklibVersion") {
+        //REMOVE below line after upgrading filepicker to latest version in arklib/arklib-debug
+        exclude group: "com.github.ARK-Builders", module: "ark-filepicker"
```
tuancoltech, line 96: These will be removed after upgrading filepicker to latest version in `arklib/arklib-debug`.

app/build.gradle
```diff
@@ -121,6 +124,9 @@ dependencies {
     implementation 'com.davemorrissey.labs:subsampling-scale-image-view-androidx:3.10.0'
     implementation 'com.github.MikeOrtiz:TouchImageView:3.1.1'
 
+    //DO NOT REMOVE THIS UNTIL filepicker is updated to latest version in arklib 0.3.2
+    implementation 'dev.arkbuilders:arkfilepicker:0.1.2-snapshot-04'
```
tuancoltech, line 128: Should remove after upgrading `filepicker` to latest version in arklib 0.3.2.

build.gradle
```diff
@@ -47,5 +47,6 @@ allprojects {
         maven { url "https://plugins.gradle.org/m2/" }
         gradlePluginPortal()
         maven { url 'https://jitpack.io' }
+        mavenLocal()
```
tuancoltech, line 50: Remove this after removing snapshot dependencies.

app/build.gradle
kirillt: Should be v0.0.6

kirillt: https://github.com/ARK-Builders/ark-components-android/packages/1930514

app/build.gradle
kirillt: But the `ark-filepicker` doesn't import `arklib-android`, it is the opposite way around.

build.gradle
kirillt: If we can safely remove jitpack.io this should be done, too.

app/build.gradle
kirillt: v0.3.3 https://github.com/ARK-Builders/arklib-android/actions/runs/6518877888

app/build.gradle
tuancoltech: @kirillt Done upgrading to v0.0.6.

build.gradle
tuancoltech: @kirillt We cannot remove `jipack.io` yet since there are still dependencies being hosted there, for instance: com.github.MikeOrtiz:TouchImageView. So let's keep it for now.

Issue #405, opened 2023-10-18
Fix Lint warnings/errors with local Analysis tool


tuancoltech:
![Screenshot 2023-10-18 at 22 27 12](https://github.com/ARK-Builders/ARK-Navigator/assets/4281969/079917cc-99af-4a6d-8d9d-13e9229cb0b3)

PR #410, opened 2023-10-23
[enhancement] Fix Lint warnings

tuancoltech:
**Fix Lint warnings:** #405

## :rocket: Summary
Fix Lint warnings including:   
* Unused namespaces in XML 
* Add task configuration avoidance by using register when define a gradle task 
* Update protected visibility in final class to private as they are effectively the same 
* Add private modifier for variables where needed 
* Upgrade kotlin version & kotlin gradle plugin to 1.8.20 
* Naming issues and others

## :framed_picture: Screenshots:

| Before |

Screenshot 2023-10-23 at 23 34 51
 
| After |

Screenshot 2023-10-23 at 23 34 10

### DIFF
--- a/app/src/main/java/dev/arkbuilders/navigator/data/stats/PlainStatsStorage.kt
+++ b/app/src/main/java/dev/arkbuilders/navigator/data/stats/PlainStatsStorage.kt
@@ -27,7 +27,7 @@ class PlainStatsStorage(
     private val index: RootIndex,
     private val preferences: Preferences,
     private val tagStorage: RootTagsStorage,
-    private val statsFlow: SharedFlow<StatsEvent>
+    statsFlow: SharedFlow<StatsEvent>
 ) : StatsStorage {
 
     private val root = index.path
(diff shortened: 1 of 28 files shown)


app/src/main/java/dev/arkbuilders/navigator/data/stats/PlainStatsStorage.kt
```diff
@@ -27,7 +27,7 @@ class PlainStatsStorage(
     private val index: RootIndex,
     private val preferences: Preferences,
     private val tagStorage: RootTagsStorage,
-    private val statsFlow: SharedFlow<StatsEvent>
+    statsFlow: SharedFlow<StatsEvent>
```
kirillt, line 30: Can it be queried directly now?

tuancoltech: @kirillt No, because `statsFlow` is now just a parameter, not a property.

Issue #412, opened 2023-11-01
Aggregated mode is broken


kirillt:
Steps:
1. Add some folders with data.
2. Make sure each folder opens without problems.
3. Click “Navigate Data” in the bottom bar.

Device used: Samsung Galaxy S10+
OS used: Android Version 12

https://github.com/ARK-Builders/ARK-Navigator/assets/581023/03514733-e8e6-4a0e-a55e-f9f6a3163b65

shakeebularsalan:
@kirillt, Verification done using APK file of 21st Oct 2023 in Samsung Galaxy S10+ (Android Version 12) and found that crash is still appearing. It is also reported as 12th bug in https://www.notion.so/ark-builders/991c95ad002a4355b3b309549b2b2323?v=33f0b6fbec3c48d5aa9ccee11f1019d1&p=3162aac1411247379376a9fceff88382&pm=s

shakeebularsalan:
@kirillt, It is reported as 12th bug in https://www.notion.so/ark-builders/991c95ad002a4355b3b309549b2b2323?v=33f0b6fbec3c48d5aa9ccee11f1019d1&p=3162aac1411247379376a9fceff88382&pm=s

kirillt:
Test data can be taken here: https://drive.google.com/drive/folders/1jYD1tDISQchVhfQqN06NG11q23ImSIE-
(There is "download all" button).

tuancoltech:
This issue was partially fixed with below 2 PRs:
https://github.com/ARK-Builders/arklib-android/pull/125
https://github.com/ARK-Builders/ARK-Navigator/pull/421

There's one more crash in `arklib` which should be address: 
https://github.com/ARK-Builders/arklib/issues/60

PR #421, opened 2023-12-05
WIP: [bug] Fix crash in Android Navigator Aggregate mode.

tuancoltech:
## :rocket: Summary
Fix crash in Android Navigator, Aggregation mode.
Issue: https://github.com/ARK-Builders/ARK-Navigator/issues/412


app/src/main/AndroidManifest.xml
```diff
@@ -7,7 +7,6 @@
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
 
     <uses-permission
-        android:requiredFeature="@string/ark_file_picker_pick"
```
tuancoltech, line 10: This is for fixing a side effect when files access permission toggle was disabled, caused by this commit: c7d14c1ac7d17ed9d0191752dbe67bd4b213467a

hieuwu (approved):

kirillt:
https://github.com/ARK-Builders/arklib-android/pull/125 merged

kirillt:
@tuancoltech am I right that we only need to bump `arklib-android` version before merging?

tuancoltech:
> @tuancoltech am I right that we only need to bump `arklib-android` version before merging?

@kirillt This PR doesn't depend on any change in `arklib-android`, so please feel free to merge it :)

kirillt:
@tuancoltech there are some unresolved references in the last build.

kirillt (approved):

app/src/main/java/dev/arkbuilders/navigator/data/stats/category/TagLabeledNStorage.kt
```diff
@@ -27,8 +27,14 @@ class TagLabeledNStorage(
...
         val storage = locateStorage()
         if (storage?.exists() == true) {
-            val json = Json.decodeFromStream<JsonTagLabeledN>(storage.inputStream())
-            tagLabeledAmount.putAll(json.data)
+            try {
+                val json = Json.decodeFromStream<JsonTagLabeledN>(
+                    storage.inputStream()
+                )
+                tagLabeledAmount.putAll(json.data)
+            } catch (exception: Exception) {
```
kirillt, line 35: It seems that just logging the exception can be harmful for the storage. We can have 2 kinds of exception here:
1. Exception during decoding the storage, i.e. reading from disk.
2. An unlikely problem during `putAll`.

In either case, the `tagLabeledAmount` mapping is incorrect. When new stats appear, the will be written to the disk, but only part of the original storage will be written. E.g. if we have 5 entries in the storage, we read 2 of them and then we have an exception. We continue execution with 2/5 entries. Then new entry is coming, so we write 3/6 to the disk. This means we lose some entries in case of the exception.

Stats storage is not that important comparing to tags or scores, but it enhances UX by sorting tags in last-used order. We should avoid losing this data if possible.

kirillt:
Another concern is that the fix is applied to `TagLabeledNStorage`, but there are also 3 other similar storages: `TagLabeledTSStorage.kt`, `TagQueriedTSStorage`, `TagQueriedNStorage`. These storages have the same bug, they should be fixed, too.

kirillt:
But the best approach would be to re-design stats storages... They were implemented during autumn 2022. At summer 2023, we've refactored storages completely and separated reusable primitives. See [FileStorage](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/FileStorage.kt) and [FolderStorage](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/data/storage/FolderStorage.kt) from arklib-android.

In fact, the data stored in stats storages is very simple key-value mapping. I think, it's possible to implement them similarly to [ScoreStorage](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/user/score/RootScoreStorage.kt). Scores storage is _id-to-integer_ mapping, and stats storages are either _tag-to-integer_ or _tag-to-timestamp_. But in any case, the code will be very similar to scores storage.

mdrlzy:
> I think, it's possible to implement them similarly to [ScoreStorage](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/dev/arkbuilders/arklib/user/score/RootScoreStorage.kt)

They are not quite like `ScoreStorage`. They can often receive new events (once per second), which will provoke overwriting of large file (there is only one for the entire root). Here we need to implement a new storage, which will have a debounce flush to disk for new entries.

I think there is no point in fixing them here; it’s better to move and refactor in arklib-android.

kirillt:
@mdrlzy good note about debouncing. We can implement such a parameter in our base storage classes. Then stats storage should be similar to score storage, just debounce parameter would differ.

kirillt:
@tuancoltech let me know if you need help with implementing `StatsStorage`

tuancoltech:
> @tuancoltech let me know if you need help with implementing `StatsStorage`

@kirillt Sure, I'll ask here in the case.

kirillt:
@tuancoltech We might need small refactoring in `arklib-android` because `BaseStorage` hard-codes `ResourceId` as the type acceptable as keys of a storage. In case of tag stats, keys belong to the `String` type.

PR #436, opened 2024-01-28
Quick build with Blake3 instead of CRC-32

kirillt:
## :rocket: Summary
CRC-32 is fast but Blake3 is cryptographic hash function.
That means that with Blake3 we won't have any id collisions.

If it isn't too slow, would be great to switch to it.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -31,7 +31,7 @@ android {
     }
 
     kotlinOptions {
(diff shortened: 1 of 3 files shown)


kirillt:
That was just an experiment. Proper build needs completion of work on `fs-index` in the Core.

PR #437, opened 2024-02-12
Improve CI by reduce dependency between steps

hieuwu:
## :rocket: Summary
Currently, total time taken to complete the whole build for each commit is ~10mins.
Build - 5 mins
Test - 2 mins (depends on build)
Lint and Klint - 2 mins (depends on test)
It takes about 7 mins before linting to be run while it is unnecessary making linting depends on test and build. 
Other than that, linting can be requested to be fixed by reviewer before they approve the PR, hence it is fine to keep linting out of build or test steps

## ✅ Solution
By removing the dependency on test, linting could be run separately, it would reduce total time it takes to complete the build from 10mins to 8mins or faster
**Result**
Screenshot 2024-02-12 at 15 22 37

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,7 +114,7 @@ jobs:
       run: ./gradlew testDebugUnitTest
 
   lint:
-    # needs: test
+    needs: test
     environment: Development
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


kirillt:
@hieuwu does this PR affect time to generate artifacts? If yes, then maybe we should depend on `build` step.

kirillt (approved):

hieuwu:
> @hieuwu does this PR affect time to generate artifacts? If yes, then maybe we should depend on `build` step.

It does not affect to any artifact uploading but all the artifacts are only available when all job are completed.

PR #439, opened 2024-03-04
Enable minify

mdrlzy:
## :rocket: Summary
Reduce apk size
`ark-navigator-arm64-v8a-apk` 34.2 MB -> 28.9 MB

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -88,8 +88,13 @@ android {
             ]
         }
         release {
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            postprocessing {
+                removeUnusedCode true
+                removeUnusedResources true
+                obfuscate false
+                optimizeCode true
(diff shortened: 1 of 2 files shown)


hieuwu (commented):
Looks good. It would be more descriptive if we keep minifyEnable property and set to true explicitly

hieuwu (approved):
Interesting number. 🚀

mdrlzy:
> Looks good. It would be more descriptive if we keep minifyEnable property and set to true explicitly

`minifyEnable` does not allow you to turn off obfuscation. I want to leave obfuscation disabled to avoid problems with json fields, reading crash logs and so on, and this is useless since the project is open source

hieuwu:
> > Looks good. It would be more descriptive if we keep minifyEnable property and set to true explicitly
> 
> `minifyEnable` does not allow you to turn off obfuscation. I want to leave obfuscation disabled to avoid problems with json fields, reading crash logs and so on, and this is useless since the project is open source

That makes sense

PR #444, opened 2024-06-21
KtLint Auto-formatting

shubertm:
## :rocket: Summary
Configure ktlint to use auto formatting after format violations checks
Issue:

## :framed_picture: Screenshots:
Provide screenshots to make it visible to reviewer if possible (Optional)

| Before | After |
| :---: | :---: |
| Screenshot_1 | Screenshot_2 |


hieuwu:
Great work. Could you please also document how to use it?

shubertm:
> Great work. Could you please also document how to use it?  

Alright, btw how are we keeping the documentation? I heard you documenting something too

kirillt:
@shubertm I think @hieuwu means some instructions in README if anything needs to be run manually by a developer.

shubertm:
> @shubertm I think @hieuwu means some instructions in README if anything needs to be run manually by a developer.
> 
> But my understanding is that Gradle will run ktlint automatically, right?

Yes, it will fix all of our formatting style violations automatically, good for new developers not to worry about code formatting errors. We did this in Retouch too. And @kirillt I was thinking if we could have this in all Kotlin repos?
@hieuwu I think there is nothing a dev is supposed to do manually.

hieuwu (approved):
Looks good. Later, do you think we can have a hook with the coding convention? Like before commit or push event, we will run check then notify in the terminal so that developer can fix their own before the code is pushed

shubertm:
> Looks good. Later, do you think we can have a hook with the coding convention? Like before commit or push event, we will run check then notify in the terminal so that developer can fix their own before the code is pushed

Yeah, earlier we had that pre-commit scripts. I will see about it. It's really important because sometimes auto format misses some

hieuwu (approved):
Looks good.
Rad LGTM

shubertm:
> Looks good.
> Rad LGTM

I think later we should add pre-commit hook for Windows too

